PR #399, opened 2023-04-23
feat(web): sandbox plugin iframe - alpha

airslice:
# Overview

This PR adds experimental sandbox iframe for NLS.
Also can be used in plugin editor with url query `?useCore&useSandbox`.

## What I've done

- Add `useExperimentalSandbox`
- Add component `SafeIFrame`


web/src/core/Crust/Plugins/PluginFrame/SafeIFrame/hooks.ts
```diff
@@ -0,0 +1,169 @@
...
+              [${JSON.stringify(autoResizeMessageKey)}]: { width, height }
+            }, "*")
+          }
+        }).observe(document.body.parentElement);
+      }
+    </script>`;
+  }, [autoResizeMessageKey]);
+
+  const srcDoc = useMemo(() => {
+    return html ? `${html}${autoResizeScript}` : "";
```
keiya01, line 160: If html is like the following, does script tag work?
```html
<DOCTYPE!>
<html>
  <body>
    <p>Hello World</p>
  </body>
</html>
```

airslice: Emmm it does not match W3C spec but most browser can still work with this. 
I updated the code to move the auto resize script into body if there's a end tag of body.

web/src/core/Crust/Plugins/PluginFrame/SafeIFrame/hooks.ts
```diff
@@ -0,0 +1,171 @@
...
+          }
+        }).observe(document.body.parentElement);
+      }
+    </script>`;
+  }, [autoResizeMessageKey]);
+
+  const srcDoc = useMemo(() => {
+    if (html === undefined) return "";
+    const htmlWithAutoResize = html?.replaceAll(/<\/body>/gi, `${autoResizeScript}</body>`);
+    return htmlWithAutoResize === html ? `${html}${autoResizeScript}` : htmlWithAutoResize;
```
keiya01, line 162: This process is a little complicated, so could you write test for this process?
For example, you can curve out this process as function in `/SafaIFrame/utils.ts`. Also you can write test for it in `/SafaIFrame/utils.test.ts`.

keiya01 (approved):

KaWaite (approved):

PR #405, opened 2023-04-25
fix(web): wrong last workspace value when multiple tabs

nina992:
# Overview
This PR is to fix returned workspace when the user open multiple tabs 
The problem was : the system saves the current workspace the user work on as the last workspace to be returned any time he uses the system , When the user opens new tab and choose anther workspace , this value is changed so when tries to do any action on the previous tab the system redirect immediately to the new one which is wrong.
## What I've done
Save The current workspace in session state , and Also save last workspace


web/src/state/index.ts
```diff
@@ -97,6 +98,29 @@ export type Workspace = {
...
+    if (value === null) {
+      return undefined;
+    }
+    return JSON.parse(value) as Workspace;
+  },
+  setItem: (key, value) => {
+    sessionStorage.setItem(key, JSON.stringify(value));
+  },
+  removeItem: function (): void {
+    throw new Error("Function not implemented.");
```
keiya01, line 113: If we log-outed application or the member is removed from workspace, it might cause something error:eyes:

nina992: if the user log-out the saved session won't be removed , but if the user log-in again and he was the same one logged last time system will retrieve the last workspace he was in and if a new user logged -in then system will take him to the default work space

nina992: @keiya01 kindly note that I tested above scenario and it is working fine

keiya01 (approved):

PR #409, opened 2023-05-04
feat(server): export dataset as csv

yk-eukarya:


server/internal/app/app.go
```diff
@@ -106,6 +106,7 @@ func initEcho(ctx context.Context, cfg *ServerConfig) *echo.Echo {
 	apiPrivate := api.Group("", private)
 	apiPrivate.POST("/graphql", GraphqlAPI(cfg.Config.GraphQL, gqldev))
 	apiPrivate.GET("/layers/:param", ExportLayer(), AuthRequiredMiddleware())
+	apiPrivate.GET("/dataset/:datasetSchemaId", ExportDataset(), AuthRequiredMiddleware())
```
rot1024, line 109:
```suggestion
	apiPrivate.GET("/datasets/:datasetSchemaId", ExportDataset(), AuthRequiredMiddleware())
```

server/internal/app/private.go
```diff
@@ -34,3 +34,25 @@ func ExportLayer() echo.HandlerFunc {
 		return c.Stream(http.StatusOK, mime, reader)
 	}
 }
+
+func ExportDataset() echo.HandlerFunc {
```
rot1024, line 38: Define this method to `adapter/http`. Usecases should be called on the adapter layer.

```
app -call-> adapter -call-> usecase -use-> domain
```

server/internal/usecase/interfaces/dataset.go
```diff
@@ -62,6 +63,7 @@ var (
 
 type Dataset interface {
 	Fetch(context.Context, []id.DatasetID, *usecase.Operator) (dataset.List, error)
+	Export(context.Context, id.DatasetSchemaID, *usecase.Operator) (io.Reader, string, error)
```
rot1024, line 66:
```suggestion
	Export(context.Context, id.DatasetSchemaID, string, *usecase.Operator) (io.Reader, string, error)
```

The name "Export" does not include the intention to export CSV, but in fact it does; why not name it ExportCSV or allow the format to be specified as an argument?

server/pkg/dataset/value.go
```diff
@@ -220,6 +220,10 @@ func (v *Value) ValuePolygon() *Polygon {
 	return nil
 }
 
+func (v *Value) String() string {
+	return v.v.String()
```
rot1024, line 224:
```suggestion
	if v == nil {
		return ""
	}
	return v.v.String()
```

server/pkg/property/value_camera.go
```diff
@@ -61,6 +61,14 @@ func (*typePropertyCamera) Validate(i interface{}) bool {
 	return ok
 }
 
+func (p *typePropertyCamera) String(i interface{}) string {
+	if !p.Validate(i) {
+		return ""
+	}
+	panic("not implemented!")
```
rot1024, line 68: PLEASE DO NOT USE PANIC IN PRODUCTION CODE

```go
return ""
```

rot1024 (changes requested):

server/pkg/property/value_typography.go
```diff
@@ -128,7 +128,8 @@ func (p *typePropertyTypography) String(i interface{}) string {
 	if !p.Validate(i) {
 		return ""
 	}
-	panic("not implemented!")
+	return ""
```
rot1024:
```go
func (p *typePropertyTypography) String(i any) string {
	return "";
}
```

`interface{}` -> `any` is recommended

server/pkg/property/value_typography.go
```diff
@@ -128,7 +128,8 @@ func (p *typePropertyTypography) String(i interface{}) string {
...
+	// Should be implemented if needed
```
rot1024, line 132: I think it is not required to implement because it is hard to represent Typography and Camera in CSV as a string.

rot1024 (approved):

PR #410, opened 2023-05-04
feat(web): support csv download for Dataset in setting page 

nina992:
# Overview
We need to support csv download for datasets listed in project setting page 
## Screenshots
image

image

## Memo
Design link : https://www.figma.com/file/bdnfDaGXGoNfUws4y7JItl/Re:Earth-UI-Gamma?node-id=13167:183453&t=mIQa6cBg5S7uC87r-1


web/src/components/molecules/Settings/Project/Dataset/DatasetList/index.stories.tsx
```diff
@@ -14,4 +14,10 @@ export default {
   component: DatasetList,
 } as Meta;
 
-export const Default = () => <DatasetList items={items} />;
+export const Default = () => (
+  <DatasetList
+    items={items}
+    removeDatasetSchema={schemaId => console.log(`Removing dataset with ID ${schemaId}`)}
+    onDownloadFile={(id, name) => console.log(`Downloading file with ID ${id} and name ${name}`)}
```
keiya01, line 21: I think you should use `@storybook/addon-actions` to mock action like the below.
```js
removeDatasetSchema={action("removeDatasetSchema")}
onDownloadFile={action("onDownloadFile")}
```

web/src/components/organisms/Settings/Project/Dataset/hooks.ts
```diff
@@ -106,6 +109,33 @@ export default (projectId: string) => {
...
+        },
+      });
+      const download = document.createElement("a");
+      download.download = name;
+      download.href = URL.createObjectURL(await res.blob());
+      download.dataset.downloadurl = [
+        "data:text/csv;charset=utf-8,",
+        download.download,
+        download.href,
+      ].join(":");
```
keiya01, line 133: Why is this necessary?

web/src/components/molecules/Settings/Project/Dataset/DatasetItem/index.tsx
```diff
@@ -13,17 +13,30 @@ export type Props = {
...
   const [isVisible, setIsVisible] = useState(false);
+  const [isDownloading, setDownloading] = useState(false);
+
   const handleRemoveDatasetSchema = useCallback(
     () => removeDatasetSchema?.(id),
     [id, removeDatasetSchema],
   );
+  const handleDownloadFile = useCallback(() => {
+    setDownloading(true);
+    onDownloadFile?.(id, name);
```
keiya01, line 37: I think `onDownloadFile` is async function. If so, it have to wait until the function is completed by using `await`.

web/src/components/molecules/Settings/Project/Dataset/DatasetList/index.stories.tsx
nina992: Sure , Done

web/src/components/organisms/Settings/Project/Dataset/hooks.ts
nina992: Without this part of the code, the download element would not have a valid downloadurl attribute.
Therefore, it is necessary to set the downloadurl attribute in order to ensure that the CSV file is downloaded correctly when the user clicks on the download link. do you want me to remove it?

web/src/components/molecules/Settings/Project/Dataset/DatasetItem/index.tsx
nina992: 'await' has no effect on the type of this expression because the onDownloadFile function is not returning a promise. so I handled async/await in onDownloadFile itself , but I added a callback here to maintain a spinning effect , would you please check

web/src/components/organisms/Settings/Project/Dataset/hooks.ts
keiya01: Hmm, I can't find specification for `data-downloadurl` attribute. I think it should remove if it works even if it is removed.

nina992: Done

keiya01 (approved):

PR #420, opened 2023-05-16
chore(web): upgrade auth0 to version 2

nina992:
# Overview
upgrade auth0 to version 2

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -93,7 +93,7 @@
     },
     "dependencies": {
         "@apollo/client": "3.7.14",
-        "@auth0/auth0-react": "1.12.1",
+        "@auth0/auth0-react": "2.1.0",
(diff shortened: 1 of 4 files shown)


web/package.json
```diff
@@ -131,6 +131,7 @@
         "jsep": "1.3.8",
         "jsonpath-plus": "7.2.0",
         "jszip": "3.10.1",
+        "latest": "^0.2.0",
```
KaWaite, line 134: Why did you add this? It seems like maybe a package that would allow a dev to setup automated checks for new upgrades on a nodejs server, but doesn't seem helpful for us.
Please remove this

nina992: done

KaWaite (approved):

PR #421, opened 2023-05-16
fix(web): add data url to componentId for better tracking of component in core

pyshx:
# Overview
This PR adds a `data?.url` to componentId on component to avoid any confusion among data with different urls.


web/src/core/engines/Cesium/Feature/index.tsx
```diff
@@ -95,7 +95,7 @@ export default function Feature({
     k: keyof AppearanceTypes,
     f?: ComputedFeature,
   ): JSX.Element | null => {
-    const componentId = `${layer.id}_${f?.id ?? ""}_${k}_${isHidden}_${
+    const componentId = `${layer.id}_${f?.id ?? ""}_${k}_${isHidden}_${data?.url ?? ""}_${
```
keiya01, line 98: I think it should use `generateIDWithMD5` to compress string like [here](https://github.com/reearth/reearth/blob/02f87bcf8f63bbc45644643adf13d8bb500dbab7/web/src/core/engines/Cesium/Feature/index.tsx#L162).

keiya01 (approved):

PR #423, opened 2023-05-16
feat(web): support nested style expression object in style evaluator

pyshx:
# Overview
This PR adds support for nested style statement to multiple degree in style expression object for style evaluator. requirement of feat was raised when the user tried to set styles for labelTypography ref: https://github.com/reearth/reearth/blob/58632a5067f7cc98764d9e942559ccc1781db60f/web/src/core/mantle/types/appearance.ts#L64


web/src/core/mantle/evaluator/simple/index.ts
```diff
@@ -76,30 +84,41 @@ export function clearAllExpressionCaches(
...
+    }
   });
 }
 
 function hasExpression(e: any): e is ExpressionContainer {
   return typeof e === "object" && e && "expression" in e;
 }
 
+function hasNonExpressionObject(v: any): boolean {
+  return typeof v === "object" && v !== null && !("expression" in v);
```
keiya01, line 119:
```suggestion
  return typeof v === "object" && v && !("expression" in v);
```
Sorry, it's really a small point, but it will confuse to me. So I think it should use the same code with `hasExpression` function.

keiya01 (approved):

web/src/core/mantle/evaluator/simple/index.ts
pyshx: Oh no, I understand.

PR #428, opened 2023-05-17
feat(web): add reearth ver support in profile header

pyshx:
# Overview
This PR introduces functionality to display the current application version in the profile header of the Reearth application, enhancing user awareness of the version they are currently utilizing.

### DIFF
--- a/web/declarations.d.ts
+++ b/web/declarations.d.ts
@@ -1,0 +1,1 @@
+declare let __APP_VERSION__: string;
--- a/web/declarations.d.ts
+++ b/web/declarations.d.ts
@@ -1,1 +1,0 @@
-declare let __APP_VERSION__: string;
--- a/web/src/components/molecules/Common/Header/profile.tsx
+++ b/web/src/components/molecules/Common/Header/profile.tsx
@@ -92,6 +92,9 @@ const HeaderProfile: React.FC<Props & Partial<LoginProps>> = ({
             <MenuListItem>
               <MenuListItemLabel icon="logout" onClick={onSignOut} text={t("Log out")} />
             </MenuListItem>
+            <MenuListItem>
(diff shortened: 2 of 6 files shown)


web/declarations.d.ts
```diff
@@ -0,0 +1 @@
+declare let __APP_VERSION__: string;
```
KaWaite, line 1: We have a 
```
declare global {
    interface Window {
  ..
...
...
}
```
in our src/config/index.ts file. Not sure it makes sense to put this there too. If it doesn't, perhaps we can move the global declaration in to this new file so that we keep things tidy.
What do you think?

web/src/components/molecules/Common/Header/profile.tsx
```diff
@@ -92,6 +92,9 @@ const HeaderProfile: React.FC<Props & Partial<LoginProps>> = ({
             <MenuListItem>
               <MenuListItemLabel icon="logout" onClick={onSignOut} text={t("Log out")} />
             </MenuListItem>
+            <MenuListItem>
+              <MenuListItemLabel text={t(`v${__APP_VERSION__}`)} />
```
KaWaite, line 96: I don't think this needs to be wrapped with the `t` function. At least for Japanese and english (and probably many languages) software stuff like v for version will not change

pyshx: Oh yeah you're right!

web/declarations.d.ts
pyshx: You're right, I may have overlooked that global class exist here. https://github.com/reearth/reearth/pull/428/commits/ff919e265894de21e1cab39955b16e487eddce9e should fix that.

KaWaite (approved):

PR #429, opened 2023-05-17
chore(web): upgrade jotai

nina992:
# Overview


web/src/state/index.ts
```diff
@@ -95,7 +96,7 @@ export type Workspace = {
   policy?: Policy | null;
 };
 
-const workspace = atomWithStorage<Workspace | undefined>("workspace", undefined);
+const workspace = atom<Workspace | undefined>(undefined);
```
nina992, line 99: I don't think it is a valid change cause I need this state to be with storage in order to fetch last workspace the user was on it before closing the browser or leaving reearth

PR #433, opened 2023-05-21
chore(server): Add REEARTH_ASSETBASEURL to /server/.env.example

aidatorajiro:
Setting AssetBaseURL config value is necessary in order to run a custom domain server.
I added the corresponding REEARTH_ASSETBASEURL environment variable to `/server/.env.example`, to avoid future confusion.

### DIFF
--- a/server/.env.example
+++ b/server/.env.example
@@ -3,7 +3,7 @@ PORT=8080
 REEARTH_DB=mongodb://localhost
 REEARTH_HOST=https://localhost:8080
 REEARTH_HOST_WEB=https://localhost:3000
-REEARTH_ASSETBASEURL=https://localhost:3000/assets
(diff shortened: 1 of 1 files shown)


server/.env.example
```diff
@@ -3,6 +3,7 @@ PORT=8080
 REEARTH_DB=mongodb://localhost
 REEARTH_HOST=https://localhost:8080
 REEARTH_HOST_WEB=https://localhost:3000
+REEARTH_ASSETBASEURL=https://localhost:3000/assets
```
rot1024, line 6:
```suggestion
REEARTH_ASSETBASEURL=https://localhost:8080/assets
```

rot1024 (changes requested):
Thank you PR. Asset will be served from backend, so default value should use `localhost:8080`

PR #438, opened 2023-05-23
fix(web): storybook v7 setup

KaWaite:
# Overview
Recent dependency upgrade task missed storybook breaking (in particular, newest Vite breaks storybook. But there were other config things not upgraded to the v7 way as well). 
This fixes the storybook setup as well as upgrades a couple old `storiesOf` stories to a more modern syntax.

## What I've done
- Added @storytelling/react-vite dependency
- Downgraded Vite
- Added storytelling dependency
- Upgraded a couple old stories files
- updated yarn storytelling script
- Because we have duplicate code in the main components as well as core/, there were conflicting stories files. Updated the names (added a `(classic)` postfix to the older, non-core ones)


keiya01 (approved):

pyshx (approved):
LGTM!
Thank You.

PR #443, opened 2023-05-24
ci: fix git hook, ESLint and CI setting to work properly

isoppp:
# Overview

This PR solves these problems.
- `yarn eslint` will freeze while it's running.
- Compared to old repository, there are some missing file
- `lint-staged` does not catch any staged files when running git-hook

## What I've done

- Add eslintignore
  - Old repository had ignored `graphql-client-api.tsx` which is the cause of the freezing problem
  - https://github.com/reearth/reearth-web/blob/main/.eslintignore#L8
- Add CI task
  - Old repository had the task in CI
  - https://github.com/reearth/reearth-web/blob/main/package.json#L18
  - https://github.com/reearth/reearth-web/blob/main/.github/workflows/ci.yml#L32
- Fix `lint-staged` setting

## How I tested

Manually run `yarn lint` and `yarn lint-staged` in my local.
CI will be displayed in this PR.

### DIFF
--- a/.github/workflows/ci_web.yml
+++ b/.github/workflows/ci_web.yml
@@ -36,7 +36,7 @@ jobs:
         - name: type
           run: yarn run type
         - name: eslint
-          run: yarn run eslint
+          run: yarn run lint
         - name: Check
           run: yarn run coverage
         - name: Send coverage report
--- a/.github/workflows/ci_web.yml
+++ b/.github/workflows/ci_web.yml
@@ -35,6 +35,8 @@ jobs:
           run: yarn install
         - name: type
           run: yarn run type
+        - name: eslint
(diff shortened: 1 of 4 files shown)


.github/workflows/ci_web.yml
```diff
@@ -35,6 +35,8 @@ jobs:
           run: yarn install
         - name: type
           run: yarn run type
+        - name: eslint
+          run: yarn run eslint
```
rot1024, line 39:
```suggestion
          run: yarn run lint
```

pyshx (approved):

rot1024 (approved):

KaWaite (approved):

PR #444, opened 2023-05-25
feat(web): Re:Earth Beta routing and basic component structure

KaWaite:
# Overview
Adding Re:Earth Beta routes and a basic component structure. 

## What I've done
- Add `/scene/:sceneId/xxxx` routing for beta
- Added Editor page and basic component structure to editor
- Added basic tab switching (Scene, Story, Widgets, Publish)
Screenshot 2023-05-26 at 12 27 40

## What I haven't done
- Any `component` level components. Just wrappers to build out the app structure
- Any finalized logic to tab switching (this can be (and probably will be) completely overhauled later.

- Route naming


web/src/app.tsx
```diff
@@ -66,7 +43,50 @@ export default function App() {
...
+                  <Route path="/settings/projects/:projectId" element={<ProjectSettings />} />
+                  <Route path="/settings/projects/:projectId/public" element={<PublicSettings />} />
+                  <Route
+                    path="/settings/projects/:projectId/dataset"
+                    element={<DatasetSettings />}
+                  />
+                  <Route
+                    path="/settings/projects/:projectId/plugins"
+                    element={<PluginSettings />}
+                  />
```
isoppp, line 83: Grouping routes like `/scene` might be more readable.

web/src/app.tsx
```diff
@@ -66,7 +43,50 @@ export default function App() {
             <Suspense fallback={<Loading />}>
               <NotificationBanner />
               <StyledRouter>
-                <AppRoutes />
+                <Routes>
```
isoppp, line 46: Separating entire `Routes` into another file might be good for maintainability and readability.

web/src/app.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { Suspense } from "react";
-import { BrowserRouter as Router, useRoutes, Navigate, useParams } from "react-router-dom";
+import { BrowserRouter as Router, Navigate, useParams, Routes, Route } from "react-router-dom";
 
+// import BetaEditor from "@reearth/beta/pages/Editor";
```
isoppp, line 4: Might be forgetting to delete

isoppp (approved):

web/src/app.tsx
KaWaite: That is true. I'll make another PR refactoring current routes just in case switching to nested routing causes unexpected issues  👍🏻

web/src/app.tsx
KaWaite: True. Removed!

web/src/app.tsx
KaWaite: This is true. I will do this as well in a separate PR!

web/src/beta/pages/Editor/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const tabs = ["scene", "story", "widgets", "publish"];
+
+const Editor: React.FC<Props> = () => {
+  const [sceneId, setSceneId] = useState<string | undefined>(undefined); // This will come from project query once beta project creation works
+  const [currentTab, setTab] = useState<Tab | undefined>();
+  const location = useLocation();
+
+  useEffect(() => {
+    const splitPathname = location.pathname.split("/");
```
keiya01, line 20: Can we use `useParams`?

web/src/beta/hooks/navigationHooks.ts
```diff
@@ -0,0 +1,21 @@
...
+    },
+    [sceneId, navigate],
+  );
+};
+
+export const useNavigationHooks = () => {
+  return {
+    useEditorNavigation,
+  };
+};
```
keiya01, line 21: BTW I'm concerned that bundling like this will increase bundle size.
It might split each hooks.

web/src/beta/pages/Editor/index.tsx
KaWaite: useParams will only get me the sceneId. But I also need the final /widgets, /story to know the tab

web/src/beta/hooks/navigationHooks.ts
KaWaite: So you think maybe its better/okay to just export everything individually, right? I think that is reasonable and maybe sticks to the initial idea better.

keiya01: Yes, I think it's better to export individual function.

web/src/beta/pages/Editor/index.tsx
keiya01: Hmm, as I thought, I think we should use params to switch tab. And redirect to 404 page if tab can't find.

web/src/beta/hooks/navigationHooks.ts
KaWaite: Done in [Update hooks](https://github.com/reearth/reearth/pull/444/commits/0fb5084f288137bf2953f2e99e7e2d6908b3a460) commit

web/src/beta/pages/Editor/index.tsx
KaWaite: [simplify Editor logic](https://github.com/reearth/reearth/pull/444/commits/c8a28b13d1ad825d2a741737e7e93bff98234ff5)

PR #447, opened 2023-05-26
refactor(web): routing

KaWaite:
# Overview
Move routing to be a service. Cleanup code and nest any routes that were applicable

### DIFF
--- a/web/src/services/routing/index.tsx
+++ b/web/src/services/routing/index.tsx
@@ -48,6 +48,7 @@ export const AppRoutes = () => {
           <Route path="publish" element={<BetaEditor />} />
         </Route>
         <Route path="settings">
+          <Route index={true} element={<Navigate to="/settings/account" />} />
           <Route path="account" element={<AccountSettings />} />
           <Route path="workspaces">
             <Route index={true} element={<WorkspaceList />} />


web/src/services/routing/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+          <Route path="preview" element={<Preview />} />
+        </Route>
+        <Route path="plugin-editor" element={<PluginEditor />} />
+        {/* classic routes - end */}
+        <Route path="scene/:sceneId" element={<BetaEditor />}>
+          <Route path="story" element={<BetaEditor />} />
+          <Route path="widgets" element={<BetaEditor />} />
+          <Route path="publish" element={<BetaEditor />} />
+        </Route>
+        <Route path="settings">
```
isoppp, line 50: Would need to add a line something like this to settings group:

```
<Route index={true} element={<Redirect to="account" />} />
```

When hit exact `/settings` then it seems not to match any route component.
Previously it redirected to `/settings/account` by https://github.com/reearth/reearth/pull/447/files#diff-663173f9ccb3067a584b11393e5923330cc0779256e7b47d6095cf11c50861e7L61

KaWaite: Nice catch
[add index to /settings](https://github.com/reearth/reearth/pull/447/commits/9982ac6f46beb90724686677666ccd20aefeac77)

isoppp (approved):
Thank you, LGTM!

PR #448, opened 2023-05-29
refactor(web): Beta tab routing

isoppp:
# Overview

Refactor tab routing logic because using useParam would be better than `string` instead of parsing pathname.
It would be the original meaning of [this comment](https://github.com/reearth/reearth/pull/444#discussion_r1206300384)

Route changes:
The URL of `scene` tab moved to `/scene/:sceneId/scene` from `/scene/:sceneId`. It might be weird but it's easier to understand.
Old `/scene/:sceneId` will be shown as 404.

## How I tested

- Navigate each tab in beta editor
- Manually go to the `/scene/:sceneId`, `/scene/:sceneId/xxxx` which shows 404 pages.


web/src/beta/pages/Editor/index.tsx
```diff
@@ -1,38 +1,28 @@
...
-      setTab("scene");
-    } else {
-      setTab(tab as Tab);
-    }
-  }, [location.pathname]);
+  const { sceneId, tab } = useParams<{ sceneId: string; tab: string }>();
+
+  if (!sceneId || !tab || !isTab(tab)) {
+    return <NotFound />;
+  }
```
isoppp, line 21: I thought the remaining original URL would be good instead of redirecting to somewhere to show 404.

KaWaite: We are already showing NotFound in the routing (with `<Route path="*" element={<NotFound />} />`

keiya01 (approved):

web/src/beta/pages/Editor/index.tsx
```diff
@@ -1,38 +1,28 @@
...
 import RightPanel from "@reearth/beta/features/RightPanel";
 import Visualizer from "@reearth/beta/features/Visualizer";
+import NotFound from "@reearth/classic/components/atoms/NotFound";
 import { styled } from "@reearth/services/theme";
 
 type Props = {};
 
-const tabs = ["scene", "story", "widgets", "publish"];
+function isTab(tab: string): tab is Tab {
+  return ["scene", "story", "widgets", "publish"].includes(tab);
```
KaWaite, line 13: We already have the ["scene, "story"....] array in navigationHooks as Tab. Since it doesn't seem to make complete sense there, can you make a types.ts file inside src/beta/ and move the Tab type to that file? 🙏🏻 Then you can import it in to the navigationHooks and this file from there

KaWaite (approved):

PR #449, opened 2023-05-29
feat(web): add aws cognito support in auth

pyshx:
# Overview
In this PR, I've integrated AWS Cognito as an authentication provider alongside the existing Auth0 implementation.

ref: https://github.com/reearth/reearth-cms/pull/727


web/src/services/auth/authProvider.tsx
```diff
@@ -0,0 +1,51 @@
...
+      </Auth0Provider>
+    ) : null;
+  }
+
+  if (authProvider === "cognito") {
+    // No specific provider needed for Cognito/AWS Amplify
+    return <CognitoWrapper>{children}</CognitoWrapper>;
+  }
+
+  return <>{children}</>; // or some default fallback
```
keiya01, line 50: Should `auth0` be used in default?

pyshx: Yes, https://github.com/reearth/reearth/blob/1b433311e75b446356e17222033eb67689bd709f/web/src/services/config/index.ts#L75

keiya01: So should we align the code like the following?
```
if(authProvider === "cognito") {
...
}
return "Use Auth0 in here"
```

pyshx: Hmm, Yeah I guess. I was mostly aiming to have it so that we might want to support something like 
Microsoft Azure Active Directory in the future.

keiya01: Hmm I see. Let's keep it.

keiya01 (approved):

PR #450, opened 2023-05-29
feat(server): support AWS Cognito auth server

yk-eukarya:


server/internal/app/config/mailer.go
```diff
@@ -0,0 +1,20 @@
...
+	Name  string
+}
+
+type SMTPConfig struct {
+	Host         string
+	Port         string
+	SMTPUsername string
+	Email        string
+	Password     string
+}
```
rot1024, line 20: They are not used. `reearthx/mailer` should be used. You can delete them.

server/internal/app/config.go
```diff
@@ -49,163 +44,21 @@ type Config struct {
...
-	AuthSrv       AuthSrvConfig
+	Auth          config.AuthConfigs
+	Auth0         config.Auth0Config
+	Cognito       config.CognitoConfig
+	AuthSrv       config.AuthSrvConfig
 	Auth_ISS      string
 	Auth_AUD      string
 	Auth_ALG      *string
 	Auth_TTL      *int
 	Auth_ClientID *string
```
rot1024, line 60: This Config struct can also be transfered to `config` package

server/internal/app/config.go
```diff
@@ -261,21 +114,41 @@ func (c Config) Print() string {
...
+		u = nil
+	}
+	return u
+}
+
+func (c Config) Auths() (res []config.AuthConfig) {
 	if ac := c.Auth0.AuthConfigs(); len(ac) > 0 {
 		res = append(res, ac...)
 	}
+	if ac := c.Cognito.AuthConfig(); ac != nil {
```
rot1024, line 137:
```go
type AuthProvider interface {
	Config() []AuthConfig
}
```

can be defined in the `config` package to simplify Auths:

```go
func (c Config) AuthConfigs() []AuthProvider {
	return []AuthProvider{c.Auth0, c.Cognito}
}

func (c Config) Auths() (res []config.AuthConfig) {
	ac := lo.FlatMap(c.AuthConfigs(), func(c AuthConfig) []AuthConfig { return c.Config() }
	// ...
}
```

rot1024 (approved):
Creating the config package is a very good idea.

server/internal/app/config/mailer.go
yk-eukarya: Done

PR #453, opened 2023-05-30
ci(web): update web codecov settings

KaWaite:
# Overview


pyshx (approved):

codecov.yml
```diff
@@ -18,3 +40,24 @@ ignore:
...
+  web-beta:
+    paths:
+      - web/src/beta
+      - web/src/services
+    carryforward: true
+    ignore:
+      - web/src/beta/utils
+  web-utils:
+    paths:
+      - web/src/beta/utils
```
keiya01, line 58:
```suggestion
      - web/src/beta/**/utils
```
I think we should check all utils. What do you think?

PR #457, opened 2023-06-01
refactor(web): storybook setup

KaWaite:
## What I've done
- Updated settings
- Added more customized theme
- Removed workaround that isn't needed anymore
- Upgraded storybook packages
- Fixed some stories.tsx files' titles to keep them organized correctly

Screenshot 2023-06-01 at 14 37 02


web/src/classic/components/molecules/Visualizer/Block/HTML/index.stories.tsx
```diff
@@ -3,6 +3,7 @@ import { Meta, Story } from "@storybook/react";
 import HTML, { Props } from ".";
 
 export default {
+  title: "molecules/Visualizer/Block/HTML",
```
pyshx, line 6: why add title here?

pyshx (approved):

web/src/classic/components/molecules/Visualizer/Block/HTML/index.stories.tsx
KaWaite: To keep it organized correctly. Without this, it becomes classic/xx/xx/xx and all other classic stories have their title explicitly defined.
In beta/ we will let storybook infer titles

pyshx: Yeah, Makes sense.

PR #463, opened 2023-06-02
chore(web): add beta editor layout

isoppp:
# Overview

Implemented the basic layout of the beta editor mainly side panels.

## What I haven't done

- storybook
- test

if the code and structure are no problem, I'm going to add stories and tests to this or another PR.
I'm thinking probably needs some changes which maybe become big in the review.

## How I tested

- manually checked on the browser so far

https://github.com/reearth/reearth/assets/16318727/abd16e9a-14e1-4e65-989a-e1e34c9d678a

https://github.com/reearth/reearth/assets/16318727/6ae81c2c-33d9-406e-9116-18985e211cab

after update branch, actually have the visualizer(but I don't plant to update above movies)
image

- file/component structure


web/src/beta/components/Resizable/README.md
```diff
@@ -0,0 +1,5 @@
+# Resizable
```
isoppp, line 1: copied from classic and edited gutter width

KaWaite: These kind of READMEs can be not copied from classic. They are from a lonnggggggg time ago and we don't want them anymore 👍🏻

isoppp: thank you! I will delete and move my comment to the component file!

web/src/beta/components/Resizable/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+`;
+
+const Gutter = styled.div`
+  background: #000000;
+  user-select: none;
+  background: ${props => props.theme.main.deepestBg};
+`;
+
+const HorizontalGutter = styled(Gutter)`
+  height: 4px;
```
isoppp, line 69: this file was almost copied from classic, then edit this gutter size ( and convert story file format to CSF3)

web/src/beta/pages/EditorPage/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+type Props = {};
+
+const EditorPage: FC<Props> = () => {
+  const { sceneId, tab } = useParams<{ sceneId: string; tab: string }>();
+
+  if (!sceneId || !tab || !isTab(tab)) {
+    return <NotFound />;
+  }
+
+  return <Editor tab={tab} sceneId={sceneId} />;
```
KaWaite, line 17: Can you combine these like below? This component will most likely become large so should keep things as tight and tidy as possible
```
  return (!sceneId || !tab || !isTab(tab)) ? <NotFound /> : <Editor tab={tab} sceneId={sceneId} />;
```

web/src/beta/features/LeftPanelScene/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+          {[...Array(100)].map((_, i) => (
+            <div key={i}>scrollable / {i}</div>
+          ))}
+        </SidePanelCard>
+      </SidePanelItem>
+    </SidePanel>
+  );
+};
+
+export default LeftPanelScene;
```
KaWaite, line 21: I feel like this can be made reuseable so you don't have to have LeftPanelScene/Story/Widget/etc.

Can't you just add `location` and `title` props (and eventually the Array(100) boilerplate will be replaced by actual content that can be passed in as a children: ReactNode) and then have a single LeftPanel and single RightPanel component?

web/src/beta/features/Navbar/index.tsx
```diff
@@ -51,9 +52,9 @@ const Navbar: React.FC<Props> = ({ sceneId, currentTab }) => {
 
 export default Navbar;
 
-const Wrapper = styled.div`
+const Wrapper = styled.div<{ theme: Theme }>`
```
KaWaite, line 55: You can access theme without this.
`useTheme()` is only needed if you need to use theme properties inline.
You can do `${({ theme }) => theme.editorNavBar.bg}` inside any styled component

web/src/beta/features/SidePanelCard/index.tsx
```diff
@@ -0,0 +1,50 @@
+import type { PropsWithChildren } from "react";
+
+import { styled, useTheme, Theme } from "@reearth/services/theme";
+
+type Props = {
+  title: string;
+};
+
+const SidePanelCard: React.FC<PropsWithChildren<Props>> = ({ title, children }) => {
```
KaWaite, line 9: Avoid PropsWithChildren
See what dan abramov (one of the main developers on React) says about it [here](https://stackoverflow.com/a/71809927/13699142)

web/src/beta/pages/EditorPage/index.tsx
keiya01: (I think we should avoid combining the condition because the condition is a little complex. If we do that, it will be looks bad to me... Also if others would like to add some condition to here, they will keep the combined way. Then It will be too complicated.)

web/src/beta/components/Resizable/hooks.ts
```diff
@@ -0,0 +1,118 @@
...
+    window.removeEventListener("touchmove", onResize);
+    window.removeEventListener("touchend", onResizeEnd);
+  }, [onResize, onResizeEnd]);
+
+  unbind = unbindEventListeners;
+
+  useEffect(() => {
+    bindEventListeners();
+    return () => unbindEventListeners();
+  });
```
keiya01, line 107: Why didn't you set deps?

web/src/beta/components/Resizable/hooks.ts
```diff
@@ -0,0 +1,118 @@
...
+  const unbindEventListeners = useCallback(() => {
+    if (typeof window === "undefined") return;
+    window.removeEventListener("mouseup", onResizeEnd);
+    window.removeEventListener("mousemove", onResize);
+    window.removeEventListener("mouseleave", onResizeEnd);
+    window.removeEventListener("touchmove", onResize);
+    window.removeEventListener("touchend", onResizeEnd);
+  }, [onResize, onResizeEnd]);
+
+  unbind = unbindEventListeners;
```
keiya01, line 102: Why do we need this line?

isoppp: I just wanted to use the same resizable as the classic and want to change only gutters size.
So didn't check it.

Took a look at it a little, but this line seems not to be able to delete easily. It might avoid an infinite loop of something.
Could I deal with it on another PR later?

web/src/beta/pages/EditorPage/index.tsx
isoppp: I thought a similar thing.
Is it no problem to change when the component gets large if needed?

web/src/beta/features/LeftPanelScene/index.tsx
isoppp: I guess after implementing logic, especially store handling stuff, the component/hooks files which have content generation might get dirty/large.
So the purpose of separating components now was that each component will be able to have complex logic or props handling in the future.

But we don't have the actual code now, once tried to fix in this [commit](https://github.com/reearth/reearth/pull/463/commits/865dc9df9def25212960d6926484bf07fb03c5cf)

web/src/beta/features/SidePanelCard/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const theme = useTheme();
```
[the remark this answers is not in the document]
isoppp: Thank you, [fixed](https://github.com/reearth/reearth/pull/463/commits/aab265ece2dfbf5cb336bda4ce9994907a8a3660)!

web/src/beta/features/Navbar/index.tsx
isoppp: Thank you, [fixed](https://github.com/reearth/reearth/pull/463/commits/aab265ece2dfbf5cb336bda4ce9994907a8a3660)!

web/src/beta/features/SidePanelCard/index.tsx
isoppp: Thank you, [fixed!](https://github.com/reearth/reearth/pull/463/commits/05369e0ce424c0ed53c608138ec2d9b1ddfed827)

web/src/beta/pages/EditorPage/index.tsx
KaWaite: I'd prefer us to keep things simpler and cleaner now, and then refactor if necessary. I don't think a single conditional that has a NotFound as a fallback is "too complicated".

If it gets more complex/large and the conditional gets hard to understand, I think the component needs a little rethinking/refactoring. 

Having a single return to look at where we can see the UI components is much simpler to read than having to search the body of the component for any/all returns...

web/src/beta/features/LeftPanelScene/index.tsx
KaWaite: Sorry the link to the commit is wrong 🙏🏻

web/src/beta/features/SidePanel/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  title: ReactNode;
+  children: ReactNode;
+  maxHeight?: CSSProperties["maxHeight"];
+};
+type Props = {
+  location: "left" | "right";
+  contents: SidePanelContent[];
+};
+
+const SidePanel: React.FC<Props> = ({ location, contents }) => {
```
KaWaite, line 16: Please keep your style the same. If you import FC directly, please always do that. If you use React.FC, please always do that

isoppp: Sorry...! I want to follow project styles (`React.FC<Props>`). Fixed in [here](https://github.com/reearth/reearth/pull/463/commits/8b7470a2039c593fcd01539bf2c9e34d6c59e633)

web/src/beta/pages/EditorPage/index.tsx
isoppp: [Fixed!](https://github.com/reearth/reearth/pull/463/commits/8b7470a2039c593fcd01539bf2c9e34d6c59e633)

KaWaite (approved):

PR #467, opened 2023-06-05
refactor(web): organize Resizable(beta) hooks

isoppp:
# Overview

Comes from [this comment](https://github.com/reearth/reearth/pull/463#discussion_r1214134840)

The previous code was a bit difficult to understand therefore make a little improvement.
It almost works the same as before.

-----

(update)

I agree and can understand what is a good solution.
But the purpose of this PR is small improvements for readability but not to implement a perfect (like attaching events only once) solution. I wanted to merely use classic components.

These are how this(both before/after PR) component is working:

### `useEffect` without deps

When `useEffect` does not have a dependencies array, this means `useEffect` runs on every render.

### How current code/this code work

a. after mount

Events are attached.

b. while resizing

When the `size` changes, then run `useEffect`.
Which means detach events and then attach events every time.
To change this, need more time.

c. after resized ( not after destroyed component )

Detach events. This is performance reason I guess because events are attached to `window` .

## How I tested

manually checked on the storybook and local Re:Earth(beta)
story: https://deploy-preview-467--reearth-web.netlify.app/storybook/?path=/story/beta-components-resizable--vertical

## Memo


web/src/beta/components/Resizable/hooks.ts
```diff
@@ -99,12 +96,14 @@ export default (
...
   }, [onResize, onResizeEnd]);
 
-  unbind = unbindEventListeners;
-
   useEffect(() => {
+    if (!isResizing) {
+      unbindEventListeners();
+      return;
+    }
     bindEventListeners();
```
keiya01, line 104:
```suggestion
    if (!isResizing) {
      return;
    }
    bindEventListeners();
```
I think this isn't necessary because the events will be unbound in the clean callback.

web/src/beta/components/Resizable/hooks.ts
```diff
@@ -99,12 +96,14 @@ export default (
...
     return () => unbindEventListeners();
-  });
+  }, [bindEventListeners, isResizing, unbindEventListeners]);
```
KaWaite, line 106: Aren't we going to get MANY bindEventListeners calls with these deps added?

My understanding is:

As the panel gets larger or smaller, the `onResize` useCallback will be updated because it has `size` as a dep. Because `onResize` gets updated, `bindEventListeners` and `unbeindEventListeners` will be updated, which in the end will make this useEffect run again, running `bindEventListeners` again which will add more event listeners to the window.

If you think that logic is incorrect, please let me know. But that is my understanding, so I think leaving this useEffect as it was with no deps is correct.

web/src/beta/components/Resizable/hooks.ts
isoppp: I added PR description. and the reason of this code is `c`.
I or the previous code want to detach events after finish resizing.

web/src/beta/components/Resizable/hooks.ts
isoppp: Added PR description.
I understand what you want to say, but working with both before/after is almost the same.
And adjusting `useEffect` is not main purpose of this PR.

KaWaite (approved):

keiya01 (approved):

PR #471, opened 2023-06-06
chore(web): add useManageSwitchState

atnagata:
# Overview
add useManageSwitchState
## How I tested
Confirmation of activation assuming multiple component (button) state use


web/src/beta/hooks/useManageSwitchState/hooks.test.tsx
```diff
@@ -0,0 +1,57 @@
+import { renderHook, act } from "@testing-library/react";
+import { expect, test } from "vitest";
+
+import useManageSwitchState, { Props } from "./hooks";
+
+type SwitchField = {
+  id: string;
+  active: boolean;
```
keiya01, line 8:
```suggestion
  name: string;
```
I think the props has `id` and `active` field in default, so could you set `name` to this field?
Also could you check if the name field is included in the fields state which is returned from the hooks?

atnagata: Added and committed.

web/src/beta/hooks/useManageSwitchState/hooks.tsx
```diff
@@ -2,6 +2,7 @@ import { useState, useCallback } from "react";
 
 type SwitchField<T> = {
   id: string;
+  name?: string;
```
keiya01, line 5: これはいらないです

KaWaite (approved):
After @keiya01 's comment is addressed and resolved you can merge!

PR #476, opened 2023-06-06
ci: fix resuable workflow

pyshx:
# Overview
This PR Attempt to fix the issue faced while using single PC for executing build-server and build-web at the same time.
ref: https://github.com/reearth/reearth/actions/runs/5187634871


.github/workflows/ci_server_build.yml
rot1024, line 1: `ci_server_build.yml` -> `build_server.yml`

.github/workflows/ci_web_build.yml
rot1024, line 1: `ci_web_build.yml` -> `build_web.yml`

rot1024 (approved):

PR #478, opened 2023-06-07
chore(web): Add stories and tests to the components related to beta-sidebar

isoppp:
# Overview

Add stories and hooks for https://github.com/reearth/reearth/pull/463

## What I haven't done

I did not add files to Resizable.
The reasons are
- There are many copied components. I guess someone needs to check all of that.
- Resizable would revamp in the future

## How I tested

run tests, check on storybook (except something related to theme)

## Memo

This branch has https://github.com/reearth/reearth/pull/467 changes so this is directing branch which is not main so far.
Therefore I can't put storybook URLs which are hosted on netlify now.


web/package.json
```diff
@@ -85,6 +85,7 @@
         "rollup": "3.22.1",
         "rollup-plugin-visualizer": "5.9.0",
         "storybook": "7.0.18",
+        "storybook-addon-react-router-v6": "^1.0.2",
```
KaWaite, line 88: Please remove the `^` and `yarn` again so we will all use the same version

web/src/beta/features/Editor/index.stories.tsx
```diff
@@ -0,0 +1,36 @@
+import { Meta, StoryObj } from "@storybook/react";
```
KaWaite, line 1: Can you write stories as shown [here](https://storybook.js.org/docs/react/writing-stories/introduction).
You can check [this component](https://github.com/reearth/reearth/blob/main/web/src/classic/components/atoms/Modal/index.stories.tsx) as an example (just don't add a `title` for beta components`

isoppp:
We will deal with these files later if needed since these files need graphQL and cesium etc in the future.

PR #479, opened 2023-06-07
fix(web): parse properties(feature) recursively before using in evaluator at core

pyshx:
This PR adds a filter to the passed feature(properties) of a feature into the evaluator before the evaluation starts. 
It has came to my attention that some of the new data contain value as object string. This fix is aimed to solve that issue.


web/src/beta/lib/core/mantle/evaluator/simple/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const generateRandomString = (len: number): string => {
+  return Array.from(window.crypto.getRandomValues(new Uint8Array(len)))
+    .map(n => usableChars[n % len])
+    .join("")
+    .toLowerCase();
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const recursiveJSONParse = (obj: any): any => {
```
keiya01, line 11:
```suggestion
export const recursiveJSONParse = (obj: unknown): object => {
```

web/src/classic/core/mantle/evaluator/simple/utils.ts
```diff
@@ -0,0 +1,30 @@
+const usableChars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+
+export const generateRandomString = (len: number): string => {
+  return Array.from(window.crypto.getRandomValues(new Uint8Array(len)))
+    .map(n => usableChars[n % len])
+    .join("")
+    .toLowerCase();
+};
+
+export const recursiveJSONParse = (obj: any): any => {
```
keiya01, line 10: Same with https://github.com/reearth/reearth/pull/479/files#r1221058941

keiya01 (approved):

PR #486, opened 2023-06-07
chore(web): design beta navBar

nina992:
# Overview

## Memo
https://www.figma.com/file/mDVeB8vBnymv2ba4RM9yTd/Re%3AEarth-2023?type=design&node-id=846-10504&t=Fk4ehf1qZ46RSr4C-0


web/src/beta/components/Button/README.md
```diff
@@ -0,0 +1,9 @@
+# Button
```
KaWaite, line 1: This file isn't necessary any more. Please delete. Its very very old

KaWaite: Please delete all of these files

web/src/beta/features/Navbar/types.ts
```diff
@@ -0,0 +1,22 @@
+import { ProjectType } from "@reearth/types";
+
+export type User = {
+  name: string;
+};
+
+export type Workspace = {
+  id?: string;
+  name?: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
KaWaite, line 10: Please don't use this unless absolutely necessary. 
Isn't the members just simply going to be `Array<User>` or something?

web/src/beta/features/Navbar/index.tsx
```diff
@@ -52,7 +82,43 @@ export default Navbar;
...
+  display: flex;
+  flex-direction: row;
+  align-items: flex-end;   
+  padding-right:24px
+  gap: 4px;
+  width: 301px;
+  height: 35px;
+`;
+
+const TabButton = styled.button<{ isSelected: boolean }>`
```
KaWaite, line 105: Can you not use components/Button as a base for the TabButtons?

web/src/beta/components/Button/README.md
nina992: done

web/src/beta/features/Navbar/index.tsx
nina992: I prefer to use this button because the Button has more styles than needed so i wanted to keep it simple rather than add  a lot of css modifications

web/src/beta/features/Navbar/types.ts
nina992: I used Uknown cause using User lead to type check errors in other places

nina992: and also because the type should match with the type existed in state and if change both may affect the classic project so I did the minimum change to avoid affect reearth classic

KaWaite: Currently, this navbar should only be used in beta (the editor), so I am not sure why you would be getting type errors, etc connected with classic

nina992: yes but state is common between both beta and classic and any change in it may affect other fields

web/src/beta/features/Navbar/index.tsx
nina992: done ( use TabButton  , move it inside navBar, modify the design)

web/src/beta/features/Navbar/types.ts
KaWaite: So can you figure out the state's type and use that here. I only said User because there is a User type in the file already so I thought maybe.
But as you say, you got a type error. So whatever the needed type is please make it that here.

web/src/beta/features/Navbar/hooks.ts
```diff
@@ -0,0 +1,122 @@
...
+    projectId: currentProject?.id ?? project?.id,
+    isPersonal: personal,
+    user,
+    currentWorkspace,
+    workspaceModalVisible,
+    handleWorkspaceModalOpen,
+    handleWorkspaceModalClose,
+    handleWorkspaceChange,
+    handleWorkspaceCreate,
+    handleLogout,
```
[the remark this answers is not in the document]
nina992: Done

web/src/beta/features/Navbar/types.ts
nina992: dear , I am using same state type but in this file there was an eslint warning , that I couldn't use any explicitly while it is not existed instate

KaWaite: Lets talk about this after daily scrum. I just want to make sure we both understand one another 👍🏻

web/src/beta/features/Navbar/hooks.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  const [currentWorkspace, setWorkspace] = useSessionWorkspace();
+  const [currentProject, setProject] = useProject();
+  const [lastWorkspace, setLastWorkspace] = useWorkspace();
+
+  const [workspaceModalVisible, setWorkspaceModalVisible] = useState(false);
+
+  const { data: workspaceData } = useGetTeamsQuery();
+  const navigate = useNavigate();
+  // const workspaces = workspaceData?.me?.teams;
```
KaWaite, line 27: Please remove

web/src/beta/features/Navbar/hooks.ts
```diff
@@ -0,0 +1,123 @@
...
+  const workspaces = useMemo(() => {
+    return workspaceData?.me?.teams?.map(({ id, name }) => ({ id, name }));
+  }, [workspaceData?.me?.teams]);
+
+  const { data } = useGetProjectBySceneQuery({
+    variables: { sceneId: sceneId ?? "" },
+    skip: !sceneId,
+  });
+  const workspaceId = data?.node?.__typename === "Scene" ? data.node.teamId : undefined;
```
KaWaite, line 36: useMemo

web/src/beta/features/Navbar/hooks.ts
```diff
@@ -0,0 +1,123 @@
...
+        ? { ...data.node.project, sceneId: data.node.id }
+        : undefined,
+    [data?.node],
+  );
+
+  const user: User = {
+    name: workspaceData?.me?.name || "",
+  };
+
+  const personal = workspaceId === workspaceData?.me?.myTeam.id;
```
KaWaite, line 49: useMemo

web/src/beta/features/Navbar/hooks.ts
nina992: done

web/src/beta/features/Navbar/hooks.ts
nina992: done

web/src/beta/features/Navbar/hooks.ts
nina992: done

KaWaite (approved):

PR #496, opened 2023-06-09
Chore(web): Add InsertionButton

atnagata:
# Overview
Add InsertionButton


web/src/beta/components/InsertionButton/index.tsx
```diff
@@ -0,0 +1,52 @@
+import styled from "@emotion/styled";
```
KaWaite, line 1: Please import styled from @reearth/services/theme

web/src/beta/components/InsertionButton/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+          padding: "2px",
+          color: "black",
+        }}
+      />
+      <Border />
+    </Box>
+  );
+};
+
+const Box = styled.div`
```
KaWaite, line 29: Can you add `pointer: cursor` to here? 🙏🏻

web/src/beta/components/InsertionButton/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+    opacity: 1;
+    transition: all 0.5s ease;
+  }
+`;
+
+const Border = styled.object`
+  height: 1px;
+  width: 100%;
+
+  background: #3b3cd0;
```
nina992, line 49: would you please use theme colors for example here 
`background:${props => props.theme.main.select};`

web/src/beta/components/InsertionButton/index.stories.tsx
```diff
@@ -0,0 +1,13 @@
...
+import InsertionButton from ".";
+
+export default {
+  component: InsertionButton,
+} as Meta;
+
+type Story = StoryObj<typeof InsertionButton>;
+
+export const Default: Story = {
+  render: () => <InsertionButton />,
```
isoppp, line 12: This line is unnecessary. If this line is written, onClick is not working.
If you want to use `render` in the future, something like `(args) => <Something {...args} />` is a basic way.

web/src/beta/components/InsertionButton/index.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export default {
+  component: InsertionButton,
+} as Meta;
+
+type Story = StoryObj<typeof InsertionButton>;
+
+export const Default: Story = {
+  args: {
+    onClick: action("onClick"),
```
isoppp, line 14: you don't need manually pass this function.
If the field name matches [this Regex](https://github.com/reearth/reearth/blob/main/web/.storybook/preview.tsx#L39), it will be automatically attached.

keiya01 (approved):

web/src/beta/components/InsertionButton/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <Icon
+        icon={"plus"}
+        style={{
+          width: "9.75px",
+          height: "9.75px",
+          background: theme.general.select,
+          borderRadius: "50%",
+          padding: "2px",
+          color: theme.general.bg.transparent,
+        }}
```
KaWaite, line 25: Please 99% of the time make use of styled components for styles.
In this case just do
```
const StyledIcon = styled(Icon)`
          width: 9.75px;
          height: 9.75px;
          background: ${({theme}) => theme.general.select};
          borderRadius: 50%;
          padding: 2px;
          color: ${({theme}) => theme.general.bg.transparent};
`;
```

KaWaite (approved):

PR #498, opened 2023-06-09
refactor(web): theme system

KaWaite:
# Overview
The theme code in reearth is the result of a few iterations and design systems over the years being partially adhered to. This Pr tries to clean up the theme system a bit and set some standards to file/folder structure.

This should perhaps be considered WIP, as I am not fully satisfied.

## What I've done
- Reorganize services/theme
  - Better organize folders
    - ex. themes are broken up, reearth themes and publish themes are split and have similar structure, etc
  - Colors are split better
- More strongly type
- Split theme between classic and beta so we don't have to worry about breaking classic styles (this includes changes to routes to wrap classic routes in classic's theme provider)

You can just focus on the code in `services/theme`. Everything else will be just import adjustments and fixes to theme/color use inside of styled components


keiya01 (approved):

pyshx (approved):
Looks good!

PR #502, opened 2023-06-14
Chore(web): Add PublishStateSwitchField

atnagata:
# Overview
Add PublishStateSwitchField


web/src/beta/components/PublishStateSwitchField/index.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+type Story = StoryObj<typeof PublishStateSwitchField>;
+
+const Wrapper: React.FC<{ children: ReactNode }> = ({ children }) => (
+  <div style={{ width: "174px", height: "24px" }}>{children}</div>
+);
+export const Default: Story = {
+  args: {
+    list: ["Unpublished", "Published"],
+    onChange: action("onchange"),
```
isoppp, line 19: you don't need manually pass this function.
If the field name matches [this Regex](https://github.com/reearth/reearth/blob/main/web/.storybook/preview.tsx#L39), it will be automatically attached.

web/src/beta/components/PublishStateSwitchField/index.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export default {
+  component: PublishStateSwitchField,
+} as Meta;
+
+type Story = StoryObj<typeof PublishStateSwitchField>;
+
+const Wrapper: React.FC<{ children: ReactNode }> = ({ children }) => (
+  <div style={{ width: "174px", height: "24px" }}>{children}</div>
+);
```
isoppp, line 15: I think this hides that the component has 100% width. It looks like 174px width component.
So it's good to put the raw component without `Wrapper`.

web/src/beta/components/PublishStateSwitchField/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+import { styled } from "@reearth/services/theme";
+
+import Icon from "../Icon";
+import Text from "../Text";
+
+type Props = {
+  list: string[];
+  selected?: string;
+  onChange?: (label: string) => void;
```
isoppp, line 11: Could you add `isActive: boolean` or something like that, and then change `StatusCircle` 's color to green if this added prop is `true` ?

isoppp, line 11:
```suggestion
  labels: string[];
  selected: string;
  onChange: (label: string) => void;
```

- `list` as plural and `label` as singular might be a bit difficult to understand.
- `selected` looks like a boolean, but actually, it's string so might be better to change
- (IMO) No one might not use this component without `selected` or `onChange`, making these fields required might be better.

web/src/beta/components/PublishStateSwitchField/index.tsx
```diff
@@ -0,0 +1,116 @@
+import { useState } from "react";
+
+import Icon from "@reearth/beta/components/Icon";
+import Text from "@reearth/beta/components/Text";
+import { styled, useTheme } from "@reearth/services/theme";
+
+const PublishStateSwitchField: React.FC = () => {
+  const labels = ["Unpublished", "Published"];
```
KaWaite, line 8: Please use i18n internationalization for text. Check classic components for examples 👍🏻

web/src/beta/components/Icon/icons.ts
```diff
@@ -37,6 +37,8 @@ import PlayRight from "./Icons/play-right.svg";
 import PlayLeft from "./Icons/play-left.svg";
 import Ellipse from "./Icons/ellipse.svg";
 
+// circle
+import PrimSphereIcon from "./Icons/primSphereIcon.svg";
```
isoppp, line 41: It seems to be unnecessary now

PR #503, opened 2023-06-14
fix(web): published page layer selection

airslice:
# Overview

Published page also needs `selectLayer` & `selectedLayerId`.

## What I've done

Only fix for classic since beta has no published page yet.


pyshx (approved):

web/src/classic/components/organisms/Published/core/hooks.ts
```diff
@@ -54,6 +58,22 @@ export default (alias?: string) => {
 
   const tags = data?.tags; // Currently no need to convert tags
 
+  const selectedLayerId = useMemo(() => {
+    return selected?.type === "layer"
+      ? { layerId: selected.layerId, featureId: selected.featureId }
+      : undefined;
+  }, [selected]);
+
+  const selectLayer = useCallback(
```
KaWaite, line 67: handleLayerSelect

web/src/classic/components/organisms/Published/core/hooks.ts
```diff
@@ -54,6 +58,22 @@ export default (alias?: string) => {
 
   const tags = data?.tags; // Currently no need to convert tags
 
+  const selectedLayerId = useMemo(() => {
```
KaWaite, line 61: You are calling this LayerId, but its layerId and featureId. Is it better to rename useSelected's value (currently `selected`) to selectedItems or something, and then call this selected?
Or....maybe something else if you can think of something that makes more sense.

KaWaite (approved):

web/src/classic/components/organisms/Published/core/hooks.ts
airslice: As we discussed we will keep it as is for classic.

PR #505, opened 2023-06-14
chore(web): Add SidePanelSectionField component

atnagata:
# Overview
Add SidePanelSectionField component


web/src/beta/components/SidePanelSectionField/index.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+      weakIcon: "",
+      select: "",
+      mask: "",
+      background: "",
+    },
+    title: "Title",
+  },
+  render: args => {
+    return <Component {...args}>{"Item"}</Component>;
+  },
```
isoppp, line 32: Could you pass `children` from `args.children` ?

web/src/beta/components/SidePanelSectionField/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+    <Field>
+      <Header onClick={() => setOpened(!opened)}>
+        <Text size="xs" color={theme.other.white} otherProperties={{ height: "16px" }}>
+          {title}
+        </Text>
+        <Icon
+          icon="arrowToggle"
+          size={12}
+          color={publishedTheme?.mainText || theme.main.text}
+          style={{ rotate: opened ? "90deg" : "180deg" }}
```
isoppp, line 26: It is better to avoid using `style` prop as much as possible.
I guess you can add a kind of wrapper and use it instead of using `style` prop

web/src/beta/components/SidePanelSectionField/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  box-sizing: border-box;
+
+  display: flex;
+  flex-direction: column;
+
+  width: 100%;
+
+  border-bottom: 1px solid ${props => props.theme.main.paleBg};
+
+  align-self: stretch;
```
isoppp, line 44: Could you remove this if you don't have any purpose? This will overwrite the parent specification.

web/src/beta/components/SidePanelSectionField/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+  );
+};
+
+const Field = styled.div`
+  box-sizing: border-box;
+
+  display: flex;
+  flex-direction: column;
+
+  width: 100%;
```
isoppp, line 40: Do you need this? This component seems that display: block is enough. ( `div` default behaviour is `block` )
Please check `width: 100%` as well

isoppp: (note) did not check modification yet, need to check on local env (?)

KaWaite (approved):

PR #507, opened 2023-06-16
feat(web): interaction mode on beta

keiya01:
# Overview

I added the interaction mode system that is necessary in VIEW3.0.

## What I've done

- Added the structure of the interaction mode system.

## What I haven't done

- Didn't add the specific feature like the sketch feature and the  multiple selection feature.

## How I tested

1. Go to Re:Earth beta project page.
2. The mode will be `default` in default.
3. Then change the mode by using `reearth.interactionMode.override("move" or "sketch" or "selection")`

These mode will work as follows.

```
- Default mode: enables `move camera` , `zoom camera`, `select entity`.
- Move mode: enables `move camera`, `zoom camera`
- Sketch mode: enables `sketch entity(not implemented yet)`, `zoom camera`
- Selection mode: enables `select multiple entities(not implemented yet)`, `zoom camera`
```


web/src/beta/lib/core/engines/Cesium/hooks.ts
```diff
@@ -370,6 +374,11 @@ export default ({
   const handleClick = useCallback(
     async (e: CesiumMovementEvent, target: RootEventTarget) => {
       mouseEventHandles.click?.(e, target);
+
+      if (!(featureFlags & MAP_FEATURE_FLAGS.SINGLE_SELECTION)) return;
+
+      // === Layer selection ===
```
KaWaite, line 380: Is this necessary? If it was on purpose, it seems like you have a couple different comment styles in this file (better if core has all the same style. Even better if ReEarth and core share the same style.).
If it was just forgotten, please remove

web/src/beta/lib/core/engines/Cesium/hooks.ts
```diff
@@ -600,6 +609,17 @@ export default ({
     [engineAPI],
   );
 
+  useEffect(() => {
+    if (!cesium.current?.cesiumElement) return;
+    const canCameraMove = !!(featureFlags & MAP_FEATURE_FLAGS.CAMERA_MOVE);
```
KaWaite, line 614: `canXxxXxx` is fine. 
But maybe `allowXxxXxxxx`, `xxxXxxxAllowed` or `xxxXxxxEnabled` is more accurate?

web/src/beta/lib/core/Visualizer/index.tsx
```diff
@@ -211,6 +218,8 @@ export default function Visualizer({
           overriddenClock={overriddenClock}
           blocks={infobox?.blocks}
           camera={camera}
+          interactionMode={interactionMode}
+          overrideInteractionMode={handleInteractionModeChange}
```
KaWaite, line 222: Is it possible to get core to adhere to our style guide?
I am not sure why we decided to (since core's development happened after the style guide was in place), but core often ignores even common React standards like callback declarations being named `handleXxxxx` and callback props being named `onXxxxxx`.
Because of this, reading the code without hover for type descriptions is hard to know what is a variable and what is a function..

KaWaite: P.S I know this is out of the scope of this PR. Just wanted to ask

web/src/beta/lib/core/Map/featureFlags.ts
```diff
@@ -0,0 +1,7 @@
+export const MAP_FEATURE_FLAGS = {
```
KaWaite, line 1: Two questions:
- Isn't this an object? Why are you calling it a map?
- If it is somehow a map, I think calling it FEATURE_FLAGS_MAP makes more sense (from an English perspective, but also if we have multiple constants directly related to feature flags, its easier to read/group like `FEATURE_FLAGS_MAP`, `FEATURE_FLAGS_OPTIONS`, etc

web/src/beta/lib/core/Crust/interactionMode.ts
```diff
@@ -0,0 +1,15 @@
+import { MAP_FEATURE_FLAGS } from "../Map/featureFlags";
+
+export type InteractionModeType = "default" | "move" | "selection" | "sketch";
+
+// If you would like to use something feature by each mode,
+// you just set the flags to enable the feature.
```
KaWaite, line 6:
```suggestion
// If you would like enable a feature in a specific mode,
// just set the feature's flag here to that mode.
```

web/src/beta/lib/core/Crust/Plugins/plugin_types.ts
```diff
@@ -23,6 +23,7 @@ import type {
 } from "@reearth/beta/lib/core/Map";
 
 import { Block } from "../Infobox";
+import { InteractionModeType } from "../types";
```
KaWaite, line 26: We have a lot of types being imported in to this file from @reearth/beta/lib/core/Map.
Does it make sense to have this type being imported from there as well? 
Or, does it make sense to move all of the above types in to this types file to consolidate types?

web/src/beta/lib/core/Map/featureFlags.ts
keiya01: This means feature flags for `core/Map`. I think we might need to define the feature flags for widget as well. But combining these flags makes the code be complicated, so I separated these implementation.

web/src/beta/lib/core/Crust/Plugins/plugin_types.ts
keiya01: I thought the mode system should be provided by `Crust`, so I defined it in here. `Map` is for abstraction for map engine, so I think it's better to place it in `Crust`.

web/src/beta/lib/core/Visualizer/index.tsx
keiya01: The reason is these functions are used as Plugin API as well. If you prefer to use `handleXxxx`, I will fix it!

web/src/beta/lib/core/engines/Cesium/hooks.ts
KaWaite: Then can you keep the style the same as elsewhere? 🙏🏻 Everywhere else is just the simple `// Comment goes here`

web/src/beta/lib/core/engines/Cesium/hooks.ts
KaWaite: Missed this comment

web/src/beta/lib/core/Map/featureFlags.ts
KaWaite: Will the flags be different? Don't we want the map and widgets to be synchronized on what features are enabled/disabled?

keiya01: No, I mean defining the flags in same place is difficult to read the code.
```
type FEATURE_FLAGS = {
  FLAG1_FOR_MAP,
  FLAG2_FOR_MAP,
  FLAG3_FOR_MAP,
  FLAG4_FOR_MAP,
  FLAG5_FOR_MAP,
  FLAG1_FOR_WIDGET,
  FLAG2_FOR_WIDGET,
  FLAG3_FOR_WIDGET,
};
```

web/src/beta/lib/core/engines/Cesium/hooks.ts
keiya01: Fixed! [Update hooks.ts](https://github.com/reearth/reearth/pull/507/commits/18c6ff9cb5e8f3263342130d618d9d2de1e8000d)

web/src/beta/lib/core/engines/Cesium/hooks.ts
keiya01: Fixed: [Update hooks.ts](https://github.com/reearth/reearth/pull/507/commits/639bfba18f19f27f407e2bb81f79e308495a0039)

web/src/beta/lib/core/Map/featureFlags.ts
keiya01: Fixed!: https://github.com/reearth/reearth/pull/507/commits/a969a5ffd1e6e1c6bbdd83446ba49c7d6d211a53

KaWaite (approved):

pyshx (approved):
LGTM, Thanks

PR #509, opened 2023-06-19
chore(web): Add SubTabButtonList component

atnagata:
# Overview
Add SubTabButtonList component

https://deploy-preview-509--reearth-web.netlify.app/storybook/?path=/story/beta-components-subtabbuttonlist--default


web/src/beta/components/SubTabButtonList/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+};
+
+const SubTabButton = styled.button<{ disabled: boolean }>`
+  background: ${({ disabled, theme }) => (disabled ? theme.main.select : "inherit")};
+  padding: 8px;
+  height: 32px;
+  border-radius: 4px 4px 0px 0px;
+  :hover {
+    background: ${({ theme }) => theme.main.select};
+    transition: all 0.5s ease;
```
isoppp, line 48: Could you move to the outside of `hover`? When something like `mouseleave` , transition is not working now.

web/src/beta/components/SubTabButtonList/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+type SwitchCustomField = SwitchField<CustomField>;
+
+export type Props = {
+  publishedTheme?: PublishTheme;
+  items: SwitchCustomField[];
+  onChange?: (id: string) => void;
+};
+const SubTabButtonList: FC<Props> = ({ publishedTheme, items, onChange }) => {
+  const theme = useTheme();
+  const { handleActivate, fields } = useManageSwitchState({ fields: items });
```
isoppp, line 20: I guess this hooks should run on the parent component.
then it will be used like `<SubButtonList items={fields} onChange={handleActive} />` and can delete `handleClick` in this component.

web/src/beta/components/SubTabButtonList/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const handleClick = (id: string) => {
+    handleActivate(id);
+    onChange?.(id);
+  };
+  return (
+    <>
+      {fields.map((item, index) => (
+        <SubTabButton
+          key={index}
+          disabled={item.active || false}
```
isoppp, line 30: this might be better.

```suggestion
          key={item.id}
          disabled={!!item?.active} // or just item.active
```

web/src/beta/components/SubTabButtonList/index.stories.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export default meta;
+
+type Story = StoryObj<typeof SubTabButtonList>;
+
+export const Default: Story = {
+  args: {
+    items: [
+      { id: "1", name: "GIS", active: true },
+      { id: "2", name: "Story", active: true },
```
KaWaite, line 17: Only one should be active at a time 🙏🏻

web/src/beta/hooks/useManageSwitchState/hooks.tsx
KaWaite, line 1: Sorry, (in this PR is okay!) can you update useManageSwitchState to keep only a single field active at one time.
Then, I think you will need to update the SubTabButtonList's story since you have but gis and story enabled in the story 🙏🏻

web/src/beta/components/SubTabButtonList/index.stories.tsx
KaWaite: If you doublecheck the useManageSwitchState it will never allow this to happen 👍🏻

atnagata: Fix completed.

web/src/beta/components/SubTabButtonList/index.tsx
atnagata: Fix completed.

web/src/beta/components/SubTabButtonList/index.tsx
atnagata: Fix completed.

web/src/beta/components/SubTabButtonList/index.tsx
atnagata: Fix completed.

isoppp: Could you move to outside this nest level? Even If you put this field to line 33, it's working as same.

atnagata: Fix completed.

isoppp (approved):

KaWaite (approved):

PR #511, opened 2023-06-20
Chore(web): Add ImageBlock

atnagata:
# Overview
Add ImageBlock
https://deploy-preview-511--reearth-web.netlify.app/storybook/?path=/story/beta-components-blocks-imageblock--default


web/src/beta/components/Blocks/ImageBlock/index.stories.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export default {
+  component: ImageBlock,
+} as Meta;
+
+type Story = StoryObj<typeof ImageBlock>;
+
+export const Default: Story = {
+  args: {
+    src: "http://placehold.it/150X150",
```
KaWaite, line 13: we already have an public directory inside of .storybook with some sample photos. Please use those so we don't have to rely on an external API 🙏🏻

web/src/beta/components/Blocks/ImageBlock/index.tsx
```diff
@@ -0,0 +1,49 @@
+import { CSSProperties, FC, ImgHTMLAttributes } from "react";
+
+import { styled } from "@reearth/services/theme";
+
+type Props = {
+  src: NonNullable<ImgHTMLAttributes<HTMLImageElement>["src"]>;
+  align?: "left" | "center" | "right"; // default center
```
KaWaite, line 7: You shouldn't need the comments here in the props.
Can you 
- remove the comments
- Inside the props destructuring(line 15) assign defaults (ie. `({src, align="center", fit="cover", .. .. . .. })`
- Lastly update the styled component's props to expect a value for fit and align always and remove the fallbacks

web/src/beta/components/Blocks/ImageBlock/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { styled } from "@reearth/services/theme";
+
+type Props = {
+  src: NonNullable<React.ImgHTMLAttributes<HTMLImageElement>["src"]>;
+  alt?: React.ImgHTMLAttributes<HTMLImageElement>["alt"];
+};
+
+const ImageBlock: React.FC<Props> = ({ src, alt }) => {
+  if (src) {
+    return (
```
KaWaite, line 11: Like the Video block , please use a ternary operator with one return 🙇🏻

web/src/beta/components/Blocks/ImageBlock/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <Wrapper>
+        <ImageBox src={src} alt={alt} />
+      </Wrapper>
+    );
+  }
+  return (
+    <BlankImageBox>
+      <Icon icon={"image"} color={"#2e2e2e"} size={32} />
```
KaWaite, line 19: Don't need { }
so `icon="image" color="#2e2e2e"`

web/src/beta/components/Blocks/ImageBlock/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+    </BlankImageBox>
+  );
+};
+
+const Wrapper = styled.div`
+  position: absolute;
```
KaWaite, line 25: Don't think you need this

web/src/beta/components/Blocks/ImageBlock/index.tsx
```diff
@@ -0,0 +1,43 @@
+import Icon from "@reearth/beta/components/Icon";
+import { styled } from "@reearth/services/theme";
+
+type Props = {
+  src: NonNullable<React.ImgHTMLAttributes<HTMLImageElement>["src"]>;
+  alt?: React.ImgHTMLAttributes<HTMLImageElement>["alt"];
```
KaWaite, line 6: You don't need to type so specifically. For some elements, where there are specific values needed, this is good.
But for something like src and alt, any string is okay, so you can just type these as `string`

isoppp (approved):

web/src/beta/components/Blocks/ImageBlock/index.tsx
FabPARKJAESUNG: The code has changed.
Above code no longer exists.

PR #514, opened 2023-06-22
fix(web): support with Authentication function support for cognito backend

pyshx:
# Overview
This PR adds support for withAuthentication function support for backend which is using cognito.

## What I've done
I've added `withAuthentication` function for cognito based authentication mode.

## How I tested
Fire-up local using cognito authentication provider and try to copy the url of the `earthEditor` or any page which required Authentication and use it in incognito mode. If it redirects to login page of cognito then its working.


nina992 (approved):

web/src/services/auth/index.tsx
```diff
@@ -18,3 +18,13 @@ export const AuthenticationRequiredPage: React.FC<{ children?: ReactNode }> = ({
...
 };
+
+export const withAuthorisation = (): ((props: any) => React.FC<any>) => {
+  const authProvider = window.REEARTH_CONFIG?.authProvider;
+
+  if (authProvider === "cognito") {
+    return withAuthenticator as unknown as (props: any) => React.FC<any>;
+  }
+
+  return withAuthenticationRequired as unknown as (props: any) => React.FC<any>;
```
KaWaite, line 29: Couple questions:
1. Do we really need to type like this? (#2 might solve this)
2. Can we encapsulate `const withAuthenticationFun = withAuthorisation();` here so that we just do `export default withAuthentication(ProjectPage);` inside each component?

web/src/services/auth/index.tsx
KaWaite, line 1: With this PR we stop using `AuthenticationRequiredPage` completely, right? So we can remove it right? 🙏🏻

pyshx: No, we `AuthenticationRequiredPage` along with `withAuthenticationFunc`. We cannot remove it.

web/src/services/auth/index.tsx
pyshx: 1. I had to, I tried to go with type like `((props: any) => JSX.Element)` which didn't worked it. Do you mean https://github.com/reearth/reearth-web/pull/2 ?
2. Something like this https://github.com/reearth/reearth/pull/514/commits/aa05523092fd7a75242d8fb8c8188ce311f03123 ?

nina992 (approved):

PR #517, opened 2023-06-23
feat(web): add an option to disable default PBR to 3dtiles and model features in NLS

keiya01:
# Overview

CustomShader: https://cesium.com/learn/cesiumjs/ref-doc/CustomShader.html
The original implementation: https://github.com/takram-design-engineering/plateau-view/blob/00fc34eda3d0b22eb7a115a948a6901f0926f860/libs/datasets/src/LambertDiffuseShader.ts#L8

|PBR|Non PBR|
|:--:|:--:|
|Screenshot 2023-06-23 at 16 39 01|Screenshot 2023-06-23 at 16 46 56|

## What I've done

- Added `pbr` property to `3dtiles` and `model` appearance.

## How I tested

1. Add 3dtiles as follow
```
reearth.layers.add({
  type: "simple",
  data: {
    type: "3dtiles",
    url: "https://assets.cms.plateau.reearth.io/assets/4f/702958-5009-4d6b-a2e0-157c7e573eb2/13100_tokyo23-ku_2022_3dtiles%20_1_1_op_bldg_13101_chiyoda-ku_lod2_no_texture/tileset.json",
  },
  "3dtiles": {
      pbr: false
  }
});
```
2. PBR should be used in default.

PR #518, opened 2023-06-23
fix(web): typing for window in beta

keiya01:
# Overview


isoppp (approved):
(note)
this PR actually fix below
- https://github.com/reearth/reearth/pull/517
- https://github.com/reearth/reearth/actions/runs/5354602622/jobs/9711849861?pr=517

PR #523, opened 2023-06-26
fix(web): revert functionality to remember last workspace opened

KaWaite:
# Overview
We were getting an infinite loop on the Dashboard that was hurting the server (probably).
The logic around remembering the previous workspace was hard to understand and maintain, so I decided to remove that implementation so we can get Re:Earth back to a working condition and then reimplement that functionality later.

### DIFF
--- a/web/src/classic/components/organisms/Authentication/RootPage/hooks.ts
+++ b/web/src/classic/components/organisms/Authentication/RootPage/hooks.ts
@@ -5,12 +5,7 @@ import { useNavigate } from "react-router-dom";
 import { useAuth, useCleanUrl } from "@reearth/services/auth";
 import { useGetTeamsQuery } from "@reearth/services/gql";
 import { useT } from "@reearth/services/i18n";
-import {
-  useWorkspace,
-  useNotification,
-  useUserId,
-  useSessionWorkspace,
-} from "@reearth/services/state";
+import { useWorkspace, useNotification, useUserId } from "@reearth/services/state";
 
 export type Mode = "layer" | "widget";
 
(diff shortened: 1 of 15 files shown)


web/src/classic/components/organisms/Authentication/RootPage/hooks.ts
```diff
@@ -19,54 +14,20 @@ export default () => {
...
-      setWorkspace(
-        workspaceId
-          ? data?.me?.teams.find(t => t.id === workspaceId) ?? data?.me?.myTeam
-          : undefined,
-      );
-      setLastWorkspace(currentWorkspace);
-
-      navigate(`/dashboard/${workspaceId}`);
-    } else {
-      setCurrentUserId(data?.me?.id);
```
isoppp, line 51: I think this function had two functionalities.

a. Handle redirect
b. Set user id if `currentUserId !== data?.me?.id`

This modification looks like losing `b`, is it merely unnecessary code?

It might be necessary when switching accounts.

web/src/classic/components/pages/Authentication/hooks.ts
```diff
@@ -45,39 +35,10 @@ export default () => {
...
-    if (currentUserId === data?.me?.id) {
-      setCurrentWorkspace(
-        workspaceId
-          ? data?.me?.teams.find(t => t.id === workspaceId) ?? data?.me?.myTeam
-          : undefined,
-      );
-      setLastWorkspace(currentWorkspace);
-      navigate(`/dashboard/${workspaceId}`);
-    } else {
-      setCurrentUserId(data?.me?.id);
```
isoppp, line 58: same as other comment

PR #527, opened 2023-06-28
fix(web):  workspace or userId being undefined

KaWaite:
# Overview


web/src/classic/components/organisms/EarthEditor/Header/hooks.ts
```diff
@@ -153,6 +163,8 @@ export default () => {
     [project, publishProjectMutation, t, setNotification],
   );
 
+  console.log("CW", currentWorkspace);
```
isoppp, line 166: Plase delete :pray:

isoppp (approved):

PR #528, opened 2023-06-28
refactor(web): handle GQL errors w useSetAtom

KaWaite:
# Overview


web/src/services/state/gqlErrorHandling.ts
```diff
@@ -0,0 +1,12 @@
+import { atom, useAtom, useSetAtom } from "jotai";
+
+// useError is needed for Apollo provider error only. Handle other errors with useNotification directly.
+type GQLError = { type?: string; message?: string };
+const error = atom<GQLError | undefined>(undefined);
+
+export const useError = () => useAtom(error);
+
+export default () => {
+  const setError = useSetAtom(error);
```
isoppp, line 10: (IMO) `export const` is simpler for me. If all `export` or file must use `default` by project rule or something, please never mind:pray:

isoppp (approved):

web/src/services/state/gqlErrorHandling.ts
KaWaite: For react custom hooks, it must be the default export or else ya, a named export would be preferable! 👍🏻

PR #531, opened 2023-06-28
refactor(web): handle cesium private property of shadow map

airslice:
# Overview

- Handle private property of cesium shadowMap as optional property.
- Throw an error if these private property is undefined.


keiya01 (approved):

pyshx (approved):
LGTM!

PR #536, opened 2023-06-29
feat(web): fetcher functions intro

pyshx:
# Overview
This introduces fetcher function in the network layer of services, with changes made in reearth-beta features.

### DIFF
--- a/web/src/beta/features/Navbar/hooks.ts
+++ b/web/src/beta/features/Navbar/hooks.ts
@@ -1,54 +1,20 @@
 import { useMemo, useCallback, useEffect, useState } from "react";
 import { useNavigate } from "react-router-dom";
 
+import { useTeamsFetcher, useProjectFetcher, useCreateTeamFetcher } from "@reearth/services/api";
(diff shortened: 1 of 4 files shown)


web/codegen.ts
```diff
@@ -0,0 +1,79 @@
...
+const rootGQLDirectory = "src/services/gql/";
+
+const rootGenerateDirectory = `${rootGQLDirectory}__gen__/`;
+
+const config: CodegenConfig = {
+  overwrite: true,
+  schema: "../server/gql/*.graphql",
+  documents: [`${rootGQLDirectory}fragments/*.ts`, `${rootGQLDirectory}queries/*.ts`],
+  generates: {
+    [rootGenerateDirectory]: {
```
pyshx, line 12: we don't need plugins any more?
["typescript", "typescript-operations", "typescript-react-apollo"],

KaWaite: True! Removed [here](https://github.com/reearth/reearth/pull/536/commits/320e2b4611444a535d9bbc3ec8e3f6d25618817a)

PR #541, opened 2023-06-30
chore(web): add empty StoryPanel

isoppp:
# Overview

Add an empty story panel and set common colors

## What I haven't done

This component will be difficult to maintain storybook so I didn't create it.

## How I tested

see editor in local

image


web/src/services/theme/reearthTheme/common/index.ts
```diff
@@ -4,7 +4,7 @@ import zIndexes, { ZIndex } from "./zIndex";
 
 export { default as GlobalStyles } from "./globalStyles";
 
-type Common = {
+export type Common = {
```
KaWaite, line 7: In classic we were exposing colors in the theme like you are doing here, but I deliberately stopped doing that so that we would use the Theme.
Colors should match [here](https://www.figma.com/file/W6nDSiCBLubcNeAT6RkOoa/Re%3AEarth-Beta-Design-system?type=design&node-id=3%3A6838&mode=design&t=k3GSw4IStkx80JEx-1), and then we use them in the Theme.
Right now you are using the common/colors like how you should use the Theme.
If there is any specific reason you are doing like this, please let me know

web/src/beta/features/Editor/index.tsx
```diff
@@ -32,7 +33,8 @@ const Editor: React.FC<Props> = ({ sceneId, tab }) => {
               {leftPanel}
             </Resizable>
           )}
-          <Center>
+          <Center hasStory={isStory}>
+            {isStory && <StoryPanel />}
```
KaWaite, line 37: I feel like the StoryPanel is in the same category as widgets and the infobox. So at the very least I think it should go in to the Visualizer, but it might make sense to put it in core (or move the widgets and infobox out of core??)
Let's discuss with @keiya01 about this next week

web/src/services/theme/reearthTheme/common/index.ts
isoppp: 1. I thought this was just type refactoring. But I noticed that [color values are exposing](https://github.com/reearth/reearth/pull/541/files#diff-968c54b183d3eacfc4e123f12973bbd93a251924731dab0211135df9c939514fR15) but type is deleting from [here](https://github.com/reearth/reearth/pull/541/files#diff-8c2d35e078b33520dcf0b2f4da48e6a00d4dae8d8174d1879d33d269223c7a62R20).
I want to delete the value as well if it's no problem...

2. we need to have another color palette which is not related to the user theme setting which means static colors. So how should I implement this if I cannot use current common colors? ( colors would add [here](https://www.figma.com/file/W6nDSiCBLubcNeAT6RkOoa/Re%3AEarth-Beta-Design-system?type=design&node-id=198-5275&mode=design&t=fpExqkMSiwMo8Kip-4) later, already used in components )

isoppp: fixed in https://github.com/reearth/reearth/pull/541/commits/25ca9526955a3580c1e7678de1cfc07f9ea67994
- Delete new color definitions
- Replace new color with hardcoded temporarily
- Fix mismatching type and value in the existing themes.

pyshx (approved):

PR #543, opened 2023-06-30
chore(web): add story page indicator

isoppp:
# Overview

Add story panel page indicator
image

storybook:
https://deploy-preview-543--reearth-web.netlify.app/storybook/?path=/story/beta-features-editor-tabs-storytelling-storypageindicator--default&args=currentPage:1;currentPageProgress:82;maxPage:9

## How I tested

on Editor and storybook


web/src/beta/features/Editor/Tabs/Storytelling/StoryPageIndicator/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  maxPage,
+  onChangePage,
+}) => {
+  return (
+    <Wrapper>
+      {Array.from({ length: maxPage }).map((_, i) => {
+        const page = i + 1;
+        const isActive = currentPage >= page;
+        const isCurrentPage = page === currentPage;
+        const progress = isCurrentPage ? currentPageProgress : isActive ? 100 : 0;
```
isoppp, line 24: I was worried about whether create a function or not and decide to not create it since it's simple enough I guess.
Let me know if you have some ideas.

web/src/beta/features/Editor/Tabs/Storytelling/StoryPanel/index.tsx
```diff
@@ -1,18 +1,39 @@
...
 const Wrapper = styled.div`
+  position: relative;
   box-sizing: border-box;
   width: 462px;
   padding: 10px 24px;
   background-color: ${({ theme }) => theme.colors.scrollBar.bg.light};
 `;
+
+const IndicatorWrapper = styled.div`
+  position: absolute;
```
KaWaite, line 35: If you position it absolutely it'll be outside of the general styles of the Wrapper that the StoryPanel will be a part of, right? 
So that'll mean if I have 10px padding set for each side of the StoryPanel, and then, for example, a Title block at the top, the title will always look too close to the indicator, no?
If the Indicator was hiding unless hovered, your approach here makes sense, but I think it'll always be visible, so not sure we should position absolutely.

web/src/beta/features/Editor/Tabs/Storytelling/StoryPageIndicator/index.tsx
KaWaite: Hmmm, no I think you are right. Since we are just displaying the prop's value a function doesn't really make sense!
Though, please wrap page, isActive, isCurrentPage and progress in useMemo. 
On scroll it will rerender A LOT, right? So we won't want more static stuff like page, isActive, etc to be re-rendered just because currentPageProgess updates

isoppp: I understand your concern and I fixed it. But the code looks a bit more complex than expected. ( Why I wrote like [this](https://github.com/reearth/reearth/pull/543/commits/ac3bc1ed67df7733e90be986044c7f763bd6c318) is memorize `progress` cannot avoid re-render since it depend on `currentPageProgress` . )

So it might be better to think about when we have actual performance issue.

web/src/beta/features/Editor/Tabs/Storytelling/StoryPanel/index.tsx
isoppp: Thank you, I did it to follow the design but misunderstood it...! Fixed https://github.com/reearth/reearth/pull/543/commits/69ff1a885ccc4eb38d70ff30eaa67b707e60748a

web/src/beta/features/Editor/Tabs/Storytelling/StoryPageIndicator/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+          <StoryPageIndicatorItem
+            key={i}
+            page={page}
+            progress={isCurrentPage ? currentPageProgress : isActive ? 100 : 0}
+            onChangePage={onChangePage}
+          />
+        );
+      })}
+    </Wrapper>
+  );
```
pyshx, line 37:
```suggestion
return (
    <Wrapper>
      {[...Array(maxPage)].map((_, i) => {
        const page = i + 1;
        const isActive = currentPage >= page;
        const progress = isActive ? 100 : 0;
        const isCurrentPage = page === currentPage;
        const indicatorProps = {
          key: i,
          page,
          progress: isCurrentPage ? currentPageProgress : progress,
          onChangePage,
        };

        return <StoryPageIndicatorItem {...indicatorProps} />;
      })}
    </Wrapper>
  );
```

Using [...Array(maxPage)] creates an array of length maxPage filled with undefined elements, which is then passed to map for iteration.

Used spread syntax (...) to pass the indicatorProps object as props to the StoryPageIndicatorItem component. This makes the code cleaner and eliminates repetitive prop assignments.

web/src/beta/features/Editor/Tabs/Storytelling/StoryPanel/index.tsx
```diff
@@ -1,18 +1,33 @@
...
-const StoryPanel: FC<Props> = () => {
-  return <Wrapper>StoryPanel</Wrapper>;
+export const StoryPanel: FC<Props> = () => {
+  return (
+    <Wrapper>
+      <StoryPageIndicator
+        currentPage={3}
+        currentPageProgress={33}
+        maxPage={6}
+        onChangePage={page => console.log(page)}
```
pyshx, line 15: Do we want to keep this log statement?

pyshx (approved):
Looks Good!

web/src/beta/features/Editor/Tabs/Storytelling/StoryPageIndicator/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  currentPageProgress: number;
+  maxPage: number;
+  onChangePage: (page: number) => void;
+};
+
+const StoryPageIndicator: FC<Props> = ({
+  currentPage,
+  currentPageProgress,
+  maxPage,
+  onChangePage,
```
KaWaite, line 18: As stated in the [style guide](https://github.com/reearth/guides/blob/main/frontend/style.md) please make sure when possible to name prop functions `onNounVerb`, like onProjectCreate, onWorkspaceDelete, etc 👍🏻

web/src/beta/features/Editor/Tabs/Storytelling/StoryPageIndicator/index.tsx
KaWaite: Ah that is true about progress, but the other variables are more static and would be nice to be memoized so that when `progress` re-renders the component they don't have to be as well.

web/src/beta/features/Editor/Tabs/Storytelling/StoryPageIndicator/StoryPageIndicatorItem/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+  page: number;
+  progress: number;
+  onChangePage: (page: number) => void;
+};
+const StoryPageIndicatorItem: FC<Props> = memo(({ progress, page, onChangePage }) => {
+  return <Indicator progress={progress} type="button" onClick={() => onChangePage(page)} />;
+});
+StoryPageIndicatorItem.displayName = "StoryPageIndicatorItem";
+
+export default memo(StoryPageIndicatorItem);
```
KaWaite, line 14: For the same reasons you mention about `progress` not being needed to be memoized, doesn't the same apply here? memoizing this component will actually be more performance heavy than not memoizing right?

web/src/beta/features/Editor/Tabs/Storytelling/StoryPageIndicator/index.tsx
isoppp: Thank you! Fixed in https://github.com/reearth/reearth/pull/543/commits/163e32207070d4421d00b2f2cf2db68571846952

web/src/beta/features/Editor/Tabs/Storytelling/StoryPanel/index.tsx
isoppp: yes, I think this is no problem since definitely replace by the end of storytelling task.

web/src/beta/features/Editor/Tabs/Storytelling/StoryPageIndicator/index.tsx
isoppp: Thank you, I applied `[...Array(maxPage)]`.

Sorry I couldn't get meaning of `eliminates repetitive prop assignments` ... :pray:

If you define the prop object first like your suggestion
- We lose type completion ( of course we can make it work by adding type )
- If return value will be bigger like `<A><B><C><D></A></B></C></D>`, we might not keep this strategy. So I feel weird.

pyshx: By using the spread syntax ({...}), we can pass all the properties of indicatorProps as separate props to the StoryPageIndicatorItem component. This eliminates the need to assign each prop individually.


In that case, you can explicitly specify the type of the indicatorProps object to maintain type checking and completion. For example:

```
const indicatorProps: StoryPageIndicatorItemProps = {
  key: i,
  page,
  progress: isCurrentPage ? currentPageProgress : progress,
  onChangePage,
};
```
By defining StoryPageIndicatorItemProps as the type for indicatorProps, you can ensure type safety and maintain proper type completion.


If the return value becomes more complex with nested components like <A><B><C><D></A></B></C></D>, it might not be suitable to use this strategy. In such cases, you can choose to keep the individual prop assignments to ensure clarity and maintainability of the code structure.


So I guess, Its upto you. Its not necessary to use them as `objectProps`..

KaWaite (approved):

PR #549, opened 2023-07-03
fix(web): selection event for published page

airslice:
# Overview

Property `selectedLayerId` `selectLayer` was not passed to `Visualizer` in published page. Therefore the selection event cannot be triggered.
We did a fix for this before, but it leads to other bugs at that time so we reverted it.
The issue was `pluginProperty` not cached in published page hooks. Which making it changes every time and leads to excute plugin code on each selection, then leads to quickjs error and variables missing, as a result we find the infobox has no data.

## What I've done

All for classic:
- Add `selectedLayerId` `selectLayer` in published page to make it support selection event.
- Memo `pluginProperty` in published page hooks.
- Fix global undefined on published page when serve on local (solution referenced index.html).

## How I tested
Tested with plateau project in both editor page and published page.
- Test selection event with dataset `VR連携機能（新大阪駅周辺地区）`
- Test infobox with dataset `建築物モデル（千代田区）`


pyshx (approved):
LGTM!

web/src/classic/components/organisms/Published/core/hooks.ts
```diff
@@ -203,6 +212,27 @@ export default (alias?: string) => {
...
+  const selectedLayerId = useMemo(() => {
+    return selected?.type === "layer"
+      ? { layerId: selected.layerId, featureId: selected.featureId }
+      : undefined;
+  }, [selected]);
+
+  const layerSelectionReason = useMemo(
+    () => (selected?.type === "layer" ? selected.layerSelectionReason : undefined),
+    [selected],
+  );
```
KaWaite, line 224: Really really small thing, but can you keep to one style. In this case, for selectedLayerId don't need return {}

KaWaite (approved):

web/src/classic/components/organisms/Published/core/hooks.ts
airslice: Updated. Thank you!

PR #551, opened 2023-07-03
chore(web): Add ColorField component

atnagata:
# Overview
Add ColorField component.
Created based on the following source.
・classic/components/molecules/EarthEditor/PropertyPane/PropertyField/ColorField


web/src/beta/components/Popup/index.tsx
```diff
@@ -0,0 +1,79 @@
+import { ReactNode, RefObject, useEffect, useRef, useCallback } from "react";
+import FocusLock from "react-focus-lock";
```
KaWaite, line 2: Can you explain why we need this? (BTW I don't think it is working the way the github page for this dependency says it should)
You can explain this to me on Slack in Japanese if its easier! 👍🏻 
I feel like we don't need this, and I'd prefer not adding a new dependency unless its VERY necessary

web/src/beta/components/BlockSettings/ColorField/styles.css
KaWaite, line 1: Can you use these inside of the styled components they affect? 🙏🏻

PR #552, opened 2023-07-03
fix(web): skip cache feature for data has updateInterval

airslice:
# Overview

 Features are been cached without data version in cache key, when feature's property been updated (one usecase is GTFS dataset for realtime bus from plateauview project) the latest values can't be applied because of the cache.

## What I've done

Skip cache featrues with data has `updateInterval` property.

## What I haven't done

Did not handle the `deleteComputedFeatures can be called with delegated data` error in this PR.

## How I tested

Test with plateauview project dataset `東京都交通局 バスロケーション情報`

## Memo

For now this solution is good for our usecase. but we may have more detailed methods to control the cache in the future.


pyshx (approved):
LGTM!

keiya01 (approved):

PR #560, opened 2023-07-05
fix(web): unexpect select undefined when select mvt layer

airslice:
# Overview

- NLS layers has a default layer {id: ''} in published page. When `lazyComputedLayerPrototype` with this layer it will throw out error and break the app.
- MVT features are not loaded in memery so finding imgery layer feature will always fail in finding the correct feature and select undefined unexpectly. This will leads to unexpect select target.

## What I've done

- Add hard coded layer id to default layer on published page.
- Remove mvt from ImageryLayerDataTypes when trying to find the feature. Keep the code in case we have other imagery layers support in the future.

## How I tested

Test with plateauview project.
- add some layer.
- add a mvt layer.
- select any normal layer.
- select mvt layer.


web/src/beta/lib/core/engines/Cesium/hooks.ts
```diff
@@ -317,7 +317,7 @@ export default ({
 
     if (!entity && selectedLayerId?.featureId) {
```
keiya01, line 318:
```suggestion
    // TODO: Support layers.selectFeature API for MVT
    if (!entity && selectedLayerId?.featureId) {
```

keiya01 (approved):

KaWaite (approved):

PR #567, opened 2023-07-10
fix(web): fix page auth bug with aws

pyshx:
# Overview
This PR attempts to fix issue observed with pages access when used with AWS backend.

## What I've done
It has been observed that sometimes the systems fails to load the env variable that is needed to identify the right authentication service we are using. I've added a new condition to tackle that and I've also made a refactor with how AuthenticationRequiredPage has been used.


web/src/services/auth/index.tsx
```diff
@@ -7,17 +7,22 @@ import { useAuthenticationRequired } from "./useAuth";
...
 
-export const withAuthorisation = (): ((props: any) => React.FC<any>) => {
+const withAuthorisation = (): ((props: any) => React.FC<any>) => {
   const authProvider = window.REEARTH_CONFIG?.authProvider;
-
   if (authProvider === "cognito") {
+    // Check if authProvider is explicitly defined as "cognito"
     return withAuthenticator as unknown as (props: any) => React.FC<any>;
+  } else if (authProvider === "auth0") {
+    // Handle the case when authProvider is undefined
```
KaWaite, line 21: This is handling if authProvider is "auth0", right?
Though I feel like the if statements speak for themselves, so not sure comments are really necessary here 🤔

pyshx: Done.

nina992 (approved):

KaWaite (approved):

PR #572, opened 2023-07-11
chore(web): panel resize functionality improvements

nina992:
# Overview
we need to update the panel resizing and add minimizing part to it 
## What I've done
- [x]  Remove max width- min width
- [x]  Minimized UI
- [x]  On resizing smaller, at 50% initial width auto minimize
- [x]  On clicking min bar, open to initial width again
## Memo
https://www.notion.so/eukarya/Update-panel-resize-ceb2b929a18949529b7dfc7343d60c6e

(added by isoppp)
[storybook](https://deploy-preview-572--reearth-web.netlify.app/storybook/?path=/story/beta-components-resizable--vertical)


web/src/beta/components/Resizable/hooks.ts
```diff
@@ -112,6 +112,9 @@ export default (
     }),
     [onResizeStart],
   );
+  const onInitializeSize = useCallback(() => {
```
KaWaite, line 115: handleResetSize

web/src/beta/components/Resizable/index.tsx
```diff
@@ -33,24 +41,50 @@ const Resizable: React.FC<Props> = ({
...
+      setMinimized(true);
+    } else {
+      setMinimized(false);
+    }
+  }, [direction, initialSize, size]);
+
+  const handleShowOriginalPanel = useCallback(() => {
+    setMinimized(false);
+    onInitializeSize();
+  }, [onInitializeSize]);
```
KaWaite, line 55: move the minimize useState in to the hooks and do this all in the original callback 🙏🏻

web/src/beta/features/Editor/index.tsx
```diff
@@ -35,8 +35,7 @@ const Editor: React.FC<Props> = ({ sceneId, projectId, workspaceId, tab }) => {
               direction="vertical"
               gutter="end"
               size={metrics.propertyMenuMinWidth}
-              minSize={metrics.propertyMenuMinWidth}
-              maxSize={metrics.propertyMenuMaxWidth}>
+              minSize={metrics.propertyMenuMinWidth / 4}>
```
KaWaite, line 38: I don't think we need `minSize` anymore since it doesn't really make sense. Instead, remove this and just pass the `initialSize` prop. (BTW, rename the prop initialSize please. We rename it in the Resizable func but not sure why the prop isn't just initialSize to begin with 👍🏻

web/src/beta/components/Resizable/index.tsx
```diff
@@ -33,24 +41,50 @@ const Resizable: React.FC<Props> = ({
...
 
-const StyledResizable = styled.div<Pick<Props, "direction" | "size">>`
+const StyledResizable = styled.div<Pick<Props, "direction" | "size" | "minSize">>`
   display: flex;
   align-items: stretch;
   flex-direction: ${({ direction }) => (direction === "vertical" ? "row" : "column")};
   width: ${({ direction, size }) => (direction === "horizontal" ? null : `${size}px`)};
   height: ${({ direction, size }) => (direction === "vertical" ? null : `${size}px`)};
   flex-shrink: 0;
+  resize: both;
```
isoppp, line 82: `resize` property would be created for another feature and it's not working since the element is `overflow: visible`. ( I read [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/resize#syntax) )
image

So It would be good to delete it.

web/src/beta/components/Resizable/index.tsx
```diff
@@ -33,24 +41,50 @@ const Resizable: React.FC<Props> = ({
...
+  min-width: ${({ direction, minSize }) =>
+    direction === "horizontal" && minSize ? `${minSize}px` : null};
+  min-height: ${({ direction, minSize }) =>
+    direction === "vertical" && minSize ? `${minSize}px` : null};
+  transition: transform 0.3s ease-in-out;
```
isoppp, line 87: I guess this element does not have transform property, so it would be unnecessary.

web/src/beta/components/Resizable/index.tsx
```diff
@@ -33,24 +41,50 @@ const Resizable: React.FC<Props> = ({
   const BottomGutter = showBottomGutter ? <HorizontalGutter {...gutterProps} /> : null;
   const LeftGutter = showLeftGutter ? <VerticalGutter {...gutterProps} /> : null;
 
+  useEffect(() => {
+    if (size <= initialSize / 2) {
```
isoppp, line 45: `minWidth / 2` instead of `initialSize / 2` might be better.
It does not work with `size: 400, minWidth: 300` props.

or please update [storybook](https://deploy-preview-572--reearth-web.netlify.app/storybook/?path=/story/beta-components-resizable--vertical) as well

web/src/beta/features/Editor/index.tsx
```diff
@@ -35,8 +35,7 @@ const Editor: React.FC<Props> = ({ sceneId, projectId, workspaceId, tab }) => {
               direction="vertical"
               gutter="end"
               size={metrics.propertyMenuMinWidth}
-              minSize={metrics.propertyMenuMinWidth}
-              maxSize={metrics.propertyMenuMaxWidth}>
```
isoppp, line 39: I guess you can delete `metrics.propertyMenuMaxWidth` from the `metrics` object

web/src/beta/components/Resizable/hooks.ts
nina992: done

web/src/beta/components/Resizable/index.tsx
```diff
@@ -33,24 +41,50 @@ const Resizable: React.FC<Props> = ({
...
+      setMinimized(true);
+    } else {
+      setMinimized(false);
+    }
```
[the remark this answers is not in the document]
nina992: done

web/src/beta/components/Resizable/index.tsx
nina992: done

web/src/beta/components/Resizable/index.tsx
nina992: yes sure 👍

web/src/beta/components/Resizable/index.tsx
nina992: We removed the minSize/maxSize from editor  based on previous discussion with kyle and based on requirement we need to compare with initialSize not minSize , For storybook  I modified it match this case.

web/src/beta/features/Editor/index.tsx
nina992: They removed now.

web/src/beta/components/Resizable/index.tsx
nina992: deleted

web/src/beta/features/Editor/index.tsx
isoppp: Sorry I meant you can delete [this line](https://github.com/reearth/reearth/blob/main/web/src/services/theme/reearthTheme/common/metrics.ts#L3).

So it's great to keep having `minSize` in this props and delete the above line.

web/src/beta/components/Resizable/index.tsx
isoppp: You can delete [this __line__](https://github.com/reearth/reearth/pull/572/files#diff-c3ec5584f7971867cb38d9a4043cad0e2895c1dddbc2a63b78812b68b8cd482dR80):pray:

```
transition: ease-in-out;
```

This means transition works 0s so the easing (`ease-in-out`) does not change anything.

web/src/beta/components/Resizable/index.tsx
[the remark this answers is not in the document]
KaWaite: I think you misunderstood the above. The above is a suggested change, so I want this only
```
 if (!minimized && size <= initialSize / 2) {
      setMinimized(true);
    }
```
instead of your current changes. Seems you copy/pasted the above code, which includes both the old changes and my recommendation.

web/src/beta/components/Resizable/index.tsx
KaWaite: You only did the first part to this comment. You moved minimized in to the hooks, but you didn't handle setting minimized to false there too. If you do that, you can delete this callback

web/src/beta/components/Resizable/index.tsx
KaWaite, line 1: We don't need min/maxSize props anymore for this. Please remove them

web/src/beta/components/Resizable/index.tsx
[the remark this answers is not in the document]
nina992: sure , my bad , removed

web/src/beta/components/Resizable/index.tsx
nina992: done

web/src/beta/components/Resizable/index.tsx
nina992: done

web/src/beta/components/Resizable/index.tsx
nina992: I removed those props from both resizable / hooks  noting that i think we should keep them in hooks for any further need

web/src/beta/features/Editor/index.tsx
nina992: done - instead I make it propertyMenuWidth to set the initial size.

KaWaite (approved):

web/src/services/theme/reearthTheme/common/metrics.ts
```diff
@@ -1,6 +1,6 @@
 const metrics = {
-  propertyMenuMinWidth: 272,
-  propertyMenuMaxWidth: 336,
+  propertyMenuWidth: 350,
```
isoppp, line 2: from design, this should be 235(left) or 286(right)

web/src/beta/components/Resizable/hooks.ts
```diff
@@ -28,22 +27,20 @@ const getPositionFromEvent = (e: React.MouseEvent | React.TouchEvent) => {
 const getDelta = (direction: Direction, deltaX: number, deltaY: number) =>
   direction === "vertical" ? deltaX : deltaY;
 
-const getSize = (size: number, delta: number, minSize?: number, maxSize?: number) => {
-  if (minSize !== undefined && size + delta < minSize) return minSize;
```
isoppp, line 32: I guess need this line to keep min-width.

web/src/beta/components/Resizable/index.tsx
isoppp: We got minWidth again in the review process, so using minWidth / 2 is better now.

web/src/services/theme/reearthTheme/common/metrics.ts
nina992: @lavalse  would you please help us here in the new beta design  i noticed that the initial size for panels is 308 and we set the min size to 200 is that ok

web/src/beta/components/Resizable/hooks.ts
isoppp: I noticed this is unnecessary

web/src/beta/components/Resizable/hooks.ts
```diff
@@ -112,6 +122,10 @@ export default (
     }),
     [onResizeStart],
   );
+  const handleResetSize = useCallback(() => {
+    setMinimized(false);
+    setSize(minSize);
```
KaWaite, line 127: Should be initialSize, not minSize here

nina992: I set it to be minSize based on your first requirement asit is applicable now after last modification

web/src/services/theme/reearthTheme/common/metrics.ts
nina992: approved by red @lavalse

web/src/beta/components/Resizable/hooks.ts
nina992: done

web/src/beta/components/Resizable/index.tsx
```diff
@@ -32,25 +30,40 @@ const Resizable: React.FC<Props> = ({
...
   display: flex;
   align-items: stretch;
   flex-direction: ${({ direction }) => (direction === "vertical" ? "row" : "column")};
   width: ${({ direction, size }) => (direction === "horizontal" ? null : `${size}px`)};
   height: ${({ direction, size }) => (direction === "vertical" ? null : `${size}px`)};
   flex-shrink: 0;
+  min-width: ${({ direction, minSize }) =>
+    direction === "horizontal" && minSize ? `${minSize}px` : null};
+  min-height: ${({ direction, minSize }) =>
+    direction === "vertical" && minSize ? `${minSize}px` : null};
```
isoppp, line 66:
```suggestion
  min-width: ${({ direction, minSize }) =>
    direction === "vertical" && minSize ? `${minSize}px` : null};
  min-height: ${({ direction, minSize }) =>
    direction === "horizontal" && minSize ? `${minSize}px` : null};
```

isoppp (approved):

PR #579, opened 2023-07-13
feat(web): new beta modal design

nina992:
# Overview
 A new design for Modal component - these changes are related to Beta project
## What I've done
- [x] Add the ability to customize Layout.
- [x] Create header and add (X) close button. 
- [x] Create footer with a divider.
- [x] check if this new design is fit wherever the modal is used before.
- [x] Modify the storybook accordingly if needed.

## screenshots 
imageimage
image
image

## Memo
https://www.figma.com/file/W6nDSiCBLubcNeAT6RkOoa/Re%3AEarth-Beta-Design-system?type=design&node-id=256-6282&mode=design


web/src/services/theme/reearthTheme/common/colors.ts
```diff
@@ -4,6 +4,19 @@ const colors = {
...
+      1: "#1e2086",
+      2: "#df3013",
+      3: "#161616",
+      4: "#393939",
+    },
+    text: {
+      main: "#E0E0E0",
+      weak: "#6F6F6F",
+      weaker: "#525252",
+    },
```
KaWaite, line 19: It seems like @lavalse is using new colours across the design. Please don't add these colors. Also, common should be COMMON. These colors are very specific to dark theme...

web/src/beta/components/Modal/index.tsx
```diff
@@ -1,54 +1,163 @@
...
+  content: ReactNode;
+  tabButton1?: ReactNode;
+  tabButton2?: ReactNode;
+};
 
 type Props = {
   className?: string;
-  title?: string;
-  size?: "sm" | "md" | "lg";
+  modalTitle?: string;
```
KaWaite, line 18: This is a modal component, with only a single title prop. `title` is enough I think

web/src/beta/components/Modal/index.tsx
```diff
@@ -1,54 +1,163 @@
...
+  size?: Size;
   button1?: ReactNode;
   button2?: ReactNode;
   children?: ReactNode;
   isVisible?: boolean;
   onClose?: () => void;
+  showSidebar?: boolean;
```
KaWaite, line 25: Like I said yesterday, this isn't necessary. If we are passing sidebarTabs, we know it needs a sidebar.

web/src/beta/components/Modal/index.tsx
```diff
@@ -1,54 +1,163 @@
...
-  size,
+  modalTitle,
+  size = "md",
   button1,
   button2,
+  children,
   isVisible,
   onClose,
-  children,
+  sidebarTabs = [],
```
KaWaite, line 38: This shouldn't have a default

web/src/beta/components/Modal/index.tsx
```diff
@@ -1,54 +1,163 @@
...
 }) => {
-  const theme = useTheme();
+  const [selectedTab, setSelectedTab] = useState<string | null>(
+    sidebarTabs.length > 0 ? sidebarTabs[0].id : null,
+  );
+
+  const handleTabChange = (tabId: string) => {
```
KaWaite, line 44: useCallback

web/src/beta/components/Modal/index.tsx
```diff
@@ -1,54 +1,163 @@
...
+        </SidebarWrapper>
+      ) : (
+        <>
+          <ContentWrapper>{children}</ContentWrapper>
+          <ButtonWrapper>
+            {button2}
+            {button1}
+          </ButtonWrapper>
+        </>
+      )}
```
KaWaite, line 85: Can you rethink this a little to simplify it. 
Basically, the modal has 4 areas:
1. (optional) title header
2. (optional) tab sidebar
3. content
4. button area

Since content and button area are consistent, it'd be nicer if we considered this Modal not as "either a tabular modal or a non-tabular modal". Instead, if we can break it down as areas (some of which are optional), it'll be cleaner I think.

web/src/beta/components/Modal/index.tsx
```diff
@@ -1,54 +1,163 @@
...
+  display: flex;
+  align-items: flex-start;
+  align-self: stretch;
+`;
+
+const NavBarWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  padding: 16px;
+  align-items: flex-start;
```
KaWaite, line 100: I think this is the default, so shouldn't need

web/src/beta/components/Modal/index.tsx
```diff
@@ -1,54 +1,163 @@
...
     </Wrapper>
   );
 };
 
-const Title = styled(Text)`
-  text-align: center;
-  margin-bottom: 24px;
+const SidebarWrapper = styled.div`
+  display: flex;
+  align-items: flex-start;
```
KaWaite, line 92: Same here

web/src/beta/components/Modal/index.tsx
```diff
@@ -1,54 +1,163 @@
...
+  align-self: stretch;
+`;
+
+const NavBarWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  padding: 16px;
+  align-items: flex-start;
+  gap: 10px;
+  align-self: stretch;
```
KaWaite, line 102: Are you trying to make sure the element takes up the full height/width? If so, flex: 1 should be good. I think stretch might have unforeseen results depending on the content

web/src/beta/components/Modal/index.tsx
```diff
@@ -1,54 +1,163 @@
...
+  border-right: 1px solid ${({ theme }) => theme.editor.weakOutline};
+`;
+
+const Tab = styled.button<{ isSelected: boolean }>`
+  display: flex;
+  padding: 4px 8px;
+  align-items: flex-start;
+  align-self: stretch;
+  border-radius: 4px;
+  background: ${({ isSelected, theme }) => (isSelected ? theme.editor.bg1 : `transparent`)};
```
KaWaite, line 112: Please only use back ticks if you are going to embed javascript values in to the string. Stick to normal "" instead

web/src/beta/components/Modal/index.tsx
```diff
@@ -1,54 +1,163 @@
...
+  align-self: stretch;
+`;
+
+const TabButtonWrapper = styled.div`
+  display: flex;
+  padding: 16px;
+  flex-direction: row;
+  justify-content: flex-end;
+  align-items: flex-start;
+  gap: 20px;
```
KaWaite, line 148: gap should be 12px based on design.

web/src/beta/components/Modal/index.tsx
KaWaite, line 1: Where appropriate, can you use the `useManageSwitchState` global hook to make sure only one tab is selected at a time.

web/src/beta/components/Modal/index.stories.tsx
```diff
@@ -25,26 +26,66 @@ const ModalContent: React.FC = () => {
...
+  id: "1",
+  label: "Tab1",
+  content: TabContent,
+  tabButton1: <Button text="Confirm" />,
+  tabButton2: <Button text="Cancel" />,
+};
+
+const sidebarTab2: SidebarTab = {
+  id: "1",
+  label: "Tab1",
```
KaWaite, line 52: Update to be consistent. Also to avoid both tabs being selected at the same time

nina992: sorry it seems last commit didn't sync right , it is already modified

web/src/services/theme/reearthTheme/common/colors.ts
nina992: as discussed , I removed those colors , and if any one is exists already I replaced with it , also I added them to common assuming that we will have same for light

web/src/beta/components/Modal/index.tsx
nina992: i keep it modalTitle cause some times the content will have Title in any further implementation for content title , but I will modified it

web/src/beta/components/Modal/index.tsx
nina992: can we discuss this point after daily scrum

web/src/beta/components/Modal/index.tsx
nina992: done

web/src/beta/components/Modal/index.tsx
nina992: All the conditions are with length so I think it is ok to keep what do you think ?

web/src/beta/components/Modal/index.tsx
nina992: done

web/src/beta/components/Modal/index.tsx
nina992: done

web/src/beta/components/Modal/index.tsx
nina992: The primary intention of the NavBarWrapper is to take up the full height of its parent container and expand vertically to accommodate the tabs. However, using flex: 1 would make the NavBarWrapper expand both horizontally and vertically, which is not desired. It would cause the NavBarWrapper to take up more width than needed, potentially affecting the layout and alignment of other elements within its parent container. 
image

web/src/beta/components/Modal/index.tsx
nina992: done

web/src/beta/components/Modal/index.tsx
nina992: would you please check I am not sure if this is the right implementation you mention

web/src/beta/components/Modal/index.tsx
nina992: done

web/src/services/theme/reearthTheme/common/colors.ts
KaWaite: Sorry can you remove those colors too. It doesn't look like you are using any of them in this PR so it would be nice to just leave theme as-is until you make your plan for the color/theme task(s) 🙏🏻

nina992: Already done

web/src/beta/components/Modal/index.tsx
nina992: The problem is that the useState in the useManageSwitchState hook is not being updated correctly,as it is capturing the initial state of TabsFields, which is an empty array. When TabsFields gets updated later, the handleActivate function does not automatically reflect that change, causing unexpected behavior. To fix this, we need to modify the useManageSwitchState hook to update the state correctly. We can achieve this by using a useEffect inside the hook to update the state whenever the fields prop changes.

web/src/beta/hooks/useManageSwitchState/hooks.tsx
nina992, line 1: The problem is that the useState in the useManageSwitchState hook is not being updated correctly,as it is capturing the initial state of TabsFields, which is an empty array. When TabsFields gets updated later, the handleActivate function does not automatically reflect that change, causing unexpected behavior. To fix this, we need to modify the useManageSwitchState hook to update the state correctly. We can achieve this by using a useEffect inside the hook to update the state whenever the fields prop changes.

web/src/beta/components/Modal/index.tsx
nina992: done

web/src/beta/components/Modal/index.tsx
nina992: done

web/src/beta/components/Modal/index.tsx
```diff
@@ -1,54 +1,140 @@
...
+  align-self: stretch;
   justify-content: flex-end;
-  margin: 20px -10px -10px 0;
+  border-top: 1px solid #525252;
+  & > :first-child {
+    order: 2;
+  }
+  & > :last-child {
+    order: 1;
+  }
```
KaWaite, line 137: I think this will be a little confusing to remember when using this Modal. Can you leave it as is please 🙇🏻

web/src/services/theme/reearthTheme/common/colors.ts
```diff
@@ -4,6 +4,10 @@ const colors = {
     white: "#FFFFFF",
     transparentBlack: "rgba(0,0,0,0.7)",
     transparentLight: "rgba(0,0,0,0.4)",
+    bg: {
+      1: "#1e2086",
+      2: "#df3013",
+    },
```
KaWaite, line 10: I don't think these are used anywhere in this PR and also background colors are 99.999% unlikely to be common across themes.
Can you remove this and leave any theme changes for your next task please 🙏🏻

KaWaite (approved):

pyshx (approved):

PR #580, opened 2023-07-14
chore(web): Add popover and basic content components

isoppp:
# Overview

Add reusable popover and menu and icons for storytelling left panel.

( The original popover idea/code comes from [official example](https://floating-ui.com/docs/react-examples) )

## What I haven't done

- A pattern that menu has a child menu
- Use of theme colors

## How I tested

- storybook

https://deploy-preview-580--reearth-web.netlify.app/storybook/?path=/story/beta-components-popover--uncontrolled
https://deploy-preview-580--reearth-web.netlify.app/storybook/?path=/story/beta-components-popover--controlled
https://deploy-preview-580--reearth-web.netlify.app/storybook/?path=/story/beta-components-popovermenucontent--default


web/src/beta/components/Popover/context.ts
```diff
@@ -0,0 +1,17 @@
+import * as React from "react";
```
KaWaite, line 1: Unless there is a specific reason you are importing like this, can you just import what you need 🙏🏻 
In the Popover/hooks.ts file as well 🙏🏻

web/src/beta/components/Popover/hooks.ts
```diff
@@ -0,0 +1,64 @@
...
+    placement,
+    open,
+    onOpenChange: setOpen,
+    whileElementsMounted: autoUpdate,
+    middleware: [
+      offset(4),
+      flip({
+        crossAxis: placement.includes("-"),
+        fallbackAxisSideDirection: "end",
+        padding: 4,
```
KaWaite, line 38: Can we make the padding a prop as well? 🙏🏻 Unless you checked all current designs and the popover is the same distance away from its parent in all current cases.

web/src/beta/components/Popover/hooks.ts
```diff
@@ -0,0 +1,64 @@
...
+  const open = controlledOpen ?? uncontrolledOpen;
+  const setOpen = setControlledOpen ?? setUncontrolledOpen;
+
+  const data = useFloating({
+    placement,
+    open,
+    onOpenChange: setOpen,
+    whileElementsMounted: autoUpdate,
+    middleware: [
+      offset(4),
```
KaWaite, line 34: Can we make the offset a prop as well? 🙏🏻 If you checked all current designs and the popover is the same distance away from its parent in all current cases this is fine though 👍🏻

web/src/beta/components/PopoverMenuContent/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+    </SRoot>
+  );
+};
+export default PopoverMenuContent;
+
+const SRoot = styled.div<Pick<Props, "width">>`
+  ${({ width }) => width && `width: ${width};`}
+
+  border-radius: 2px;
+  border: 1px solid #525252;
```
nina992, line 61: Is it possible to use theme colors instead of plain color

isoppp: We plan to update the theme object with new design system variables.
So I'm thinking using current theme variables will make it more difficult to replace.

web/src/beta/components/Popover/context.ts
isoppp: Thank you. Fixed in https://github.com/reearth/reearth/pull/580/commits/eadef081850ff82f435d2e4acd4b59a5553653ad .

web/src/beta/components/Popover/hooks.ts
isoppp: Thank you. Fixed in https://github.com/reearth/reearth/pull/580/commits/eadef081850ff82f435d2e4acd4b59a5553653ad .

nina992 (approved):

KaWaite (approved):

PR #583, opened 2023-07-19
chore(web): enable visualizer explicit rendering

airslice:
# Overview

Enable explicit rendering for beta.

Cesium [demo](https://sandcastle.cesium.com/?src=Scene%20Rendering%20Performance.html) here.

## What I've done

- Enable `requestRenderMode`
- Add an eventListener to postRender event to trigger requestRender.
- Add a `RequestingRenderMode` Ref as a flag to trigger render when scene property or layer changes.
  - `-1`: force render
  - `0`: disable
  - `1`: render once
- Add `scene.render.debugFramePerSecond: boolean` to help observe the render behavior.   

## How I tested

- 3d tiles
- Animated model
- DnD

### Enable debug frame per second:
```js
reearth.scene.overrideProperty({
    render:{
        debugFramePerSecond: true,
    }
})
```

### Add 3d tiles: 
```js
const layerId = reearth.layers.add({
  type: "simple",
  data: {
    type: "3dtiles",
    url: "https://assets.cms.plateau.reearth.io/assets/11/6d05db-ed47-4f88-b565-9eb385b1ebb0/13100_tokyo23-ku_2022_3dtiles%20_1_1_op_bldg_13101_chiyoda-ku_lod1/tileset.json",
  },
  "3dtiles": {
  }
});
setTimeout(()=>{
  reearth.camera.flyTo(layerId, {duration: 0});
},10);
```

### Override 3dtiles layer property
Note: Test this AFTER the 3d tiles fully loaded, since its load will also trigger render.
```js
reearth.layers.override(layerId, {
    "3dtiles":{
        color: 'red'
    }
})
```

### Add model
```js
const modelLayerId = reearth.layers.add({
  type: "simple",
  data: {
    type: "geojson",
    value: {
      type: "Feature",
      geometry: {
        type: "Point",
        coordinates: [139.753214, 35.684891, 1], // Japan Tokyo Chiyoda-ku
      },
    },
  },
  model: {
    url: "https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/Fox/glTF-Binary/Fox.glb",
    scale: 1000,
    specularEnvironmentMaps: "https://cesium.com/public/SandcastleSampleData/kiara_6_afternoon_2k_ibl.ktx2",
  },
});
setTimeout(()=>{
  reearth.camera.flyTo(modelLayerId, {duration: 0});
},10);
```

### Override scene property to enable animation
```js
reearth.scene.overrideProperty({
    timeline:{
        animation: true,
    },
    render:{
        debugFramePerSecond: true,
    }
})
```

### Override scene property
```js

reearth.scene.overrideProperty({
    default:{
        skybox: false,
    },
    render:{
        debugFramePerSecond: true,
    }
})
```

### CZML
```js
const czml = reearth.layers.add({
  type: "simple",
  data: {
    type: "czml",
    url: "https://assets.cms.plateau.reearth.io/assets/a3/37e424-3420-41b5-9166-b7160cf539ca/uc22-015_hololab_fieldwork_.czml",
  },
});
setTimeout(()=>{
  reearth.camera.flyTo(czml, {duration: 0});
},5000);
```

### Clipping box
```js
const clippingbox = reearth.layers.add({
  type: "simple",
  data: {
    type: "3dtiles",
    url: "https://assets.cms.plateau.reearth.io/assets/11/6d05db-ed47-4f88-b565-9eb385b1ebb0/13100_tokyo23-ku_2022_3dtiles%20_1_1_op_bldg_13101_chiyoda-ku_lod1/tileset.json",
  },
  "3dtiles": {
      experimental_clipping: {
        width: 100,
        height: 100,
        length: 100,
        pointOutlineWidth: 1,
        useBuiltinBox: true,
        visible: true,
        coordinates: [139.753214, 35.684891, 1]
      }
  },
    box: {
      outlineColor: "#ffffff",
      activeOutlineColor: "#0ee1ff",
      outlineWidth: 1,
      draggableOutlineWidth: 10,
      draggableOutlineColor: "rgba(14, 225, 255, 0.5)",
      activeDraggableOutlineColor: "rgba(14, 225, 255, 1)",
      fillColor: "rgba(255, 255, 255, 0.1)",
      axisLineColor: "#ffffff",
      axisLineWidth: "#ffffff",
      pointFillColor: "rgba(255, 255, 255, 0.3)",
      pointOutlineColor: "rgba(14, 225, 255, 0.5)",
      activePointOutlineColor: "rgba(14, 225, 255, 1)",
      width: 100,
      height: 100,
      length: 100,
    }
});
setTimeout(()=>{
  reearth.camera.flyTo(clippingbox, {duration: 0});
},10);
```


web/src/beta/lib/core/engines/Cesium/hooks.ts
```diff
@@ -718,6 +723,36 @@ export default ({
...
+    const viewer = cesium.current?.cesiumElement;
+    if (!requestingRender?.current || !viewer || viewer.isDestroyed()) return;
+    viewer.scene.requestRender();
+    requestingRender.current = false;
+  }, [requestingRender]);
+
+  useEffect(() => {
+    const viewer = cesium.current?.cesiumElement;
+    if (!viewer || viewer.isDestroyed()) return;
+    viewer.scene.postUpdate.addEventListener(explicitRender);
```
keiya01, line 737: We can use [this hooks](https://github.com/reearth/reearth/blob/0b8919e8abeb529f13d3066c24a54cbf7436ff9c/web/src/beta/lib/core/engines/Cesium/hooks/useSceneEvent.ts#L37) here.

airslice: Updated. Thank you!

airslice: Sorry i revert this part as we discussed the scene event hook can't be used here.

pyshx (approved):
LGTM!

web/src/beta/lib/core/Map/Layers/hooks.ts
```diff
@@ -561,6 +565,11 @@ export default function useHooks({
     setOverridenLayers(layers => layers.filter(l => !deleted.includes(l.id)));
   }, [atomMap, layers, layerMap, lazyLayerMap, setOverridenLayers, showLayer]);
 
+  useEffect(() => {
+    if (!requestingRenderMode || requestingRenderMode.current === -1) return;
```
keiya01, line 569: I think we should use the constant variable for this value. For example
```
const FORCE_REQUEST_RENDER  = -1;
const DISABLE_REQUEST_RENDER  = 0;
const ENABLE_REQUEST_RENDER  = 1;
```

web/src/beta/lib/core/engines/Cesium/hooks.ts
```diff
@@ -740,6 +747,42 @@ export default ({
...
+    if (requestingRenderMode.current === 1) {
+      requestingRenderMode.current = 0;
+    }
+  }, [requestingRenderMode]);
+
+  useEffect(() => {
+    const viewer = cesium.current?.cesiumElement;
+    if (!viewer || viewer.isDestroyed()) return;
+    return viewer.scene.postUpdate.addEventListener(explicitRender);
+  }, [explicitRender]);
```
keiya01, line 764: I think we should wrap the `explicitRender` function with the ref. Because `postUpdate` event is invoked in every frame, but the request render will be executed after one frame if `postUpdate` is listened depends on updating the `explicitRender` function.
```suggestion
  const explicitRenderRef = useRef();
  useEffect(() => {
    explicitRenderRef.current = explicitRender;
  }, [explicitRender]);
  useEffect(() => {
    const viewer = cesium.current?.cesiumElement;
    if (!viewer || viewer.isDestroyed()) return;
    return viewer.scene.postUpdate.addEventListener(() => {
      explicitRenderRef.current?.();
    });
  }, []);
```

web/src/beta/lib/core/Map/Layers/hooks.ts
```diff
@@ -561,6 +565,11 @@ export default function useHooks({
...
+    requestingRenderMode.current = 1;
```
keiya01, line 570: It's really simple rather than the way I expected!!

web/src/beta/lib/core/engines/Cesium/hooks.ts
airslice: Yes it seems better this way. Thank you!

web/src/beta/lib/core/Map/Layers/hooks.ts
airslice: Yes i think this can work well for most of the usecases, but need to take care of none-resource related async updates. In this case the requested render could happen before all async task finishes, and that's why i added a request render from engine API after compute features finish for 3d tiles.

web/src/beta/lib/core/Map/Layers/hooks.ts
airslice: Updated!

web/src/beta/lib/core/Map/Layers/hooks.ts
keiya01: I remembered that. I think we need to invoke requestRender after the following lines
- Resource feature(This attaches the style to CZML or KML's features)
  - https://github.com/reearth/reearth/blob/ad4ae6df063b73641bd001a70d8767c53f10e42a/web/src/beta/lib/core/engines/Cesium/Feature/Resource/index.tsx#L81C51-L81C51
  - https://github.com/reearth/reearth/blob/ad4ae6df063b73641bd001a70d8767c53f10e42a/web/src/beta/lib/core/engines/Cesium/Feature/Resource/index.tsx#L146-L148
- Also I guess the clipping box is not working
  -  I think we need to invoke `requestRender` after every `updateTrs` function like [here](https://github.com/reearth/reearth/blob/ad4ae6df063b73641bd001a70d8767c53f10e42a/web/src/beta/lib/core/engines/Cesium/Feature/Box/hooks/box.ts#L81).
  - But if you are not sure about this part, I can handle it. Let me know:pray:

airslice: Updated.
Note there's some perfermance issue around clipping box.

keiya01 (approved):

PR #590, opened 2023-07-26
fix(web): gpx parsing not working

pyshx:
# Overview
This PR fixes issue observed while Parsing `gpx` data and In previous implementation utilizing 3rd party parser was leading to missing of data and mismatch of issue. This refactor should address the issue and make `gpx` parsing functioning.

### DIFF
--- a/web/src/classic/core/mantle/data/gpx.ts
+++ b/web/src/classic/core/mantle/data/gpx.ts
@@ -1,4 +1,3 @@
-import { XMLParser, X2jOptionsOptional } from "fast-xml-parser";
 import type { LineString, Position, GeoJsonProperties } from "geojson";
 
 import type { Data, DataRange, Feature } from "../types";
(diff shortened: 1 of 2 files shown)


web/src/classic/core/mantle/data/gpx.ts
```diff
@@ -14,111 +13,547 @@ export async function fetchGPXfile(
...
+        text: getElementValue(linkElement, "text"),
+        type: getElementValue(linkElement, "type"),
+      };
+    }
+
+    // Parse and store all points in the route
+    const routePoints = Array.from(routeElement.querySelectorAll("rtept"));
+    for (const routePoint of routePoints) {
+      const point: Point = {
+        latitude: parseFloat(routePoint.getAttribute("lat") ?? ""),
```
keiya01, line 169: `parseFloat("")` returns NaN. I think we should handle this case with more safer way.

keiya01 (approved):

web/src/classic/core/mantle/data/gpx.ts
pyshx: https://github.com/reearth/reearth/pull/590/commits/686efefaeb0ea8739da9f6bb27464449c4f54428 Should handle it.

web/src/beta/lib/core/mantle/data/gpx.ts
```diff
@@ -14,111 +13,547 @@ export async function fetchGPXfile(
...
+    }
+  }
+
+  // Parse and store all waypoints
+  const waypoints = Array.from(output.xml.querySelectorAll("wpt"));
+  for (const waypoint of waypoints) {
+    const point: Waypoint = {
+      name: getElementValue(waypoint, "name"),
+      symbol: getElementValue(waypoint, "sym"),
+      latitude: parseFloat(waypoint.getAttribute("lat") ?? "") || 0,
```
keiya01, line 110:
```suggestion
      latitude: parseFloat(waypoint.getAttribute("lat") ?? 0),
```
I think we should do like this to avoid confusing.

pyshx: Oh Yeah, Apologies for my ignorance.

keiya01 (approved):

nina992 (approved):

PR #553, opened 2023-07-03
feat(server): support storytelling

yk-eukarya:

PR #592, opened 2023-07-26
chore(web): hook up with story page API

isoppp:
# Overview

currently including [this PR](https://github.com/reearth/reearth/pull/553)'s changes to work before it's done. So please ignore `/server` changes for now.

Added these features by hooking up with API
- initial story creation when scene is created in dashboard
- story page
  - addition
  - deletion
  - DnD

## What I haven't done

~~- DnD is buggy now ( maybe API side ).~~ fixed.
- Duplication feature since API is not implemented yet.

## How I tested

- Create a new beta project and open editor and story tab then play with it
  - I guess we can try on only local(?) 
- Modified component can check on storybook
  - https://deploy-preview-592--reearth-web.netlify.app/storybook/?path=/story/beta-features-editor-tabs-story-pageindicator--default
  - https://deploy-preview-592--reearth-web.netlify.app/storybook/?path=/story/beta-features-editor-tabs-story-sidepanel-contentpage--default


web/src/beta/features/Editor/tabs/story/PageIndicator/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { FC } from "react";
```
isoppp, line 1: Fixed this because the animation is weird when the number of pages increases by more than 1 at a time.

web/src/services/gql/fragments/story.ts
```diff
@@ -0,0 +1,12 @@
+import { gql } from "@apollo/client";
+
+export const storyFragment = gql`
+  fragment StoryFragment on Story {
+    id
+    title
+    pages {
+      id
```
KaWaite, line 8: You have id here and on the storyPageFragment. Is that on purpose, or we can remove here?

web/src/services/api/storytellingApi.ts
```diff
@@ -0,0 +1,132 @@
...
+    },
+    [createStoryMutation, setNotification, t],
+  );
+
+  const [createStoryPageMutation] = useMutation<
+    CreateStoryPageMutation,
+    MutationCreateStoryPageArgs
+  >(CREATE_STORY_PAGE);
+
+  const createStoryPage = useCallback(
```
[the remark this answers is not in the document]
KaWaite: We are using useXXXX to mirror apollo query and gql standards around their hooks (ie useQuery, useLazyQuery, useMutation, etc).
I get what you mean, and maybe in the future we will refactor them to more standard functional names (ie handleXxxXxx), but for now can you name the mutation and query callbacks to the same standard? 🙏🏻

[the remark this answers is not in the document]
KaWaite: Also, there is a type `MutationReturn` for mutation callbacks. An example for project creation's return type is: `Promise<MutationReturn<Partial<Project>>>`. Please add that so we can make sure we are returning status correctly as well as the appropriate object

web/src/classic/components/organisms/Dashboard/hooks.ts
```diff
@@ -192,26 +194,43 @@ export default (workspaceId?: string) => {
...
         });
-        if (scene.errors) {
+        setModalShown(false);
+        return;
+      }
+
+      if (data.projectType === "beta") {
+        const story = await createStory({
+          sceneId: scene.data?.createScene?.scene.id,
+          title: "Default",
```
KaWaite, line 214: Can you i18n the title? 🙏🏻

web/src/beta/pages/Page.tsx
```diff
@@ -55,16 +55,24 @@ const PageWrapper: React.FC<Props> = ({ sceneId, projectId, workspaceId, childre
   ) : (
     <>
       <GlobalModal />
-      {childrenWithProps}
+      {renderItem({
+        sceneId,
+        projectId: currentProjectId,
+        workspaceId: currentWorkspaceId,
+        stories: scene?.stories ?? [],
```
KaWaite, line 62: The only tab that cares about stories is the storytelling page, but here you are treating it like all pages care and need them. 
Can you setup a useStorytellingQuery inside of the API, and fetch the stories, etc at that level? 
Anyways, it'll use cached values since scene has already been queried, but it saves us from long props lists with things that are only related to specific tabs, pages, components, etc. and keeps intent obvious and simpler 🙏🏻

web/src/beta/features/Editor/tabs/story/StoryPanel/index.tsx
```diff
@@ -1,19 +1,36 @@
...
+  }, [onPageSelect, selectedPage, selectedStory]);
   return (
     <Wrapper>
-      <PageIndicator
-        currentPage={3}
-        currentPageProgress={33}
-        maxPage={6}
-        onPageChange={page => console.log(page)}
-      />
+      {!!selectedStory?.pages?.length && (
```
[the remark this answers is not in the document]
KaWaite: Don't think we need to show this indicator with a single story

web/src/beta/features/Editor/tabs/story/SidePanel/ContentStory/index.tsx
```diff
@@ -53,7 +53,7 @@ const ContentStory: React.FC<Props> = ({
           }}
           renderItem={item => {
             return (
-              <PageItemWrapper pageCount={item.index + 1} isSwipable={item.index % 2 === 0}>
+              <PageItemWrapper pageCount={item.index + 1} isSwipeable={item.index % 2 === 0}>
```
KaWaite, line 56: We can just do `item.index === 0` here, since triple equals is strict.
As reference:
```
undefined == 0 // true
undefined === 0 // false
```

web/src/classic/components/organisms/Dashboard/hooks.ts
KaWaite: This doesn't seem to be resolved

web/src/beta/pages/Page.tsx
isoppp: We talked that it is good to consider in another PR since the idea needs experiment/investigation.
So mark as resolved :bow:

web/src/services/gql/fragments/story.ts
isoppp: Thank you, fixed in https://github.com/reearth/reearth/pull/592/commits/4437e0d873eeea14c5b1c7b68c992f05e75296e0

web/src/classic/components/organisms/Dashboard/hooks.ts
isoppp: didn't fix yet in that time..:pray:
Fixed in https://github.com/reearth/reearth/pull/592/commits/276e09ef78d3e5a27d40fc80c5f268c06d6ae306

web/src/services/api/storytellingApi.ts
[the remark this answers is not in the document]
isoppp: Fixed in https://github.com/reearth/reearth/pull/592/commits/eeb17278c4f7968cb68359f6ba95860bd67567fc and https://github.com/reearth/reearth/pull/592/commits/0f7915e19785bc1a86ce38ba497755b00ac2e20a

Added errors to mutation return since we sometimes need it.
https://github.com/reearth/reearth/commit/0f7915e19785bc1a86ce38ba497755b00ac2e20a#diff-1d80e33480c523e4a6f8dcff37487486759567b683aa0dcab664f5d077f8b28cR11

KaWaite (approved):

PR #597, opened 2023-07-27
fix(web): create workspace  beta modal design

nina992:
# Overview
Create workspace modal design had a side empty space so in this pr I fixed it 
image

so the  fixed one is : 
image


web/src/beta/components/Modal/index.tsx
```diff
@@ -67,7 +67,7 @@ const Modal: React.FC<Props> = ({
       isVisible={isVisible}
       title={title}
       onClose={onClose}>
-      <InnerWrapper>
+      <InnerWrapper hasTabs={tabs.length > 0}>
```
KaWaite, line 70: Instead of this, just add `flex: 1` to ContentWrapper 👍🏻 

Also, can you adjust the margins on the WorkspaceCreationModal's `FormInputWrapper` so the spacing looks more appropriate 🙏🏻

nina992: done

airslice (approved):

PR #610, opened 2023-07-31
feat(server): storytelling blocks CRUD

yk-eukarya:


server/internal/usecase/interactor/storytelling.go
```diff
@@ -468,17 +474,182 @@ func (i *Storytelling) RemovePageLayer(ctx context.Context, inp interfaces.PageL
...
+	}
+
+	err = i.storytellingRepo.Save(ctx, *story)
+	if err != nil {
+		return nil, nil, nil, -1, err
+	}
+
+	tx.Commit()
+	return story, page, block, 1, err
 }
```
pyshx, line 545: Looks like you can separate the transaction-related functions and make them as a higher-order function which accepts the function to execute as an argument.

Something like
```
func (i *Storytelling) executeInTransaction(ctx context.Context, toExecute func(context.Context) error) error {
	tx, err := i.transaction.Begin(ctx)
	if err != nil {
		return err
	}
	ctx = tx.Context()
	defer func() {
		if err2 := tx.End(ctx); err == nil && err2 != nil {
			err = err2
		}
	}()
	err = toExecute(ctx)
	if err == nil {
		tx.Commit()
	}
	return err
}
```
You can refactor other functions (RemoveBlock, MoveBlock) in a similar way, extracting out common parts to helper methods or functions and using the executeInTransaction higher-order function to encapsulate the transaction handling code.

pyshx (changes requested):

server/internal/usecase/interactor/storytelling.go
yk-eukarya: Will work on this on different PR

pyshx (approved):
LGTM, Thanks.

KaWaite (approved):

PR #612, opened 2023-08-01
chore(web): fix several issues around visualizer render

airslice:
# Overview

Fix several issues around rendering.

## What I've done

1. About override layer:
   - Reverted. Nothing changed. 
 
2. About explicit rendering:
   - Fix rendering on Resource loaded.
   - Fix rendering around clock.

3. About the engine:
   - Fix `flyTo` missing options.


web/src/beta/lib/core/Map/Layers/utils.ts
```diff
@@ -5,11 +5,10 @@ export const deepAssign = <O extends Record<string, any>>(obj: O, src: O) => {
     Object.entries({ ...src, ...obj })
       .map(([k, v]): [string, any] | undefined => {
         const srcV: unknown = src[k];
-        if (Object.hasOwn(src, k) && (srcV === undefined || srcV === null)) {
+        if (Object.hasOwn(src, k) && srcV === undefined) {
```
airslice, line 8: When override with explicit value `null`, i think it's better to set value to `null` then ignore it. 
With this change we can clear some property value, otherwise if the layer has some property when adding, we can not clear it by override.

web/src/beta/lib/core/Map/Layers/utils.ts
```diff
@@ -5,11 +5,10 @@ export const deepAssign = <O extends Record<string, any>>(obj: O, src: O) => {
...
           return undefined;
         }
-
-        if (srcV === undefined) {
+        if (srcV === undefined || (isObject(srcV) && Object.keys(srcV).length === 0)) {
```
airslice, line 11: When override with emtpy object like `marker: {}` it will compute correctly for `overriddenLayer`, but compute incorrectly for `computedLayer` (the `marker` also become {}).

airslice: I revert this part since the previous change already solved the problem. And this is changing the behavior so better keep it as is.

web/src/beta/lib/core/Map/Layers/utils.ts
keiya01: Can you add test for this case?:pray:

web/src/beta/lib/core/engines/Cesium/index.tsx
```diff
@@ -185,6 +185,8 @@ const Cesium: React.ForwardRefRenderFunction<EngineRef, EngineProps> = (
         useWebVR={!!property?.default?.vr || undefined}
         light={light}
         useDepthPicking={false}
+        maximumRenderTimeChange={0}
+        requestRenderMode={true}
```
keiya01, line 189: Why should we set these property to scene as well?

airslice: Removed. I think it's unnecessary since we already set it on Viewer.

web/src/beta/lib/core/Map/Layers/utils.ts
airslice: Sorry i realize we don't use `null` as appearance property values. So i revert this part too. When override values to reset or clear i think it's better to use empty string or empty array base on the tyep definiation.

keiya01 (approved):

pyshx (approved):
Thanks.

PR #616, opened 2023-08-02
chore(server): add spacing value type with padding and margin UI support

yk-eukarya:


server/pkg/property/value_spaceing.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	v := Spacing{}
+	if err := mapstructure.Decode(i, &v); err == nil {
+		return v, true
+	}
+	return nil, false
+}
+
+func (*typePropertySpacing) V2I(v interface{}) (interface{}, bool) {
```
pyshx, line 49: The `V2I` method seems to not be performing any useful operation.

pyshx (approved):

server/pkg/property/value_spaceing.go
yk-eukarya: this method exists because I'm implementing the `TypeProperty` interface and I'm forced to implement this method.

rot1024 (approved):

PR #617, opened 2023-08-02
chore(server): add JSON schema to built in manifest

yk-eukarya:


pyshx (approved):

server/pkg/builtin/manifest.yml
```diff
@@ -1,3 +1,4 @@
+$schema: manifest.schema.json
```
rot1024, line 1: refering to `schemas/plugin_manifest.json` is not good? https://github.com/reearth/reearth/blob/main/server/schemas/plugin_manifest.json

rot1024 (approved):

rot1024 (approved):

PR #625, opened 2023-08-07
fix(web): import builtin plugin from external url

keiya01:
# Overview

I tried to fix the unsafe builtin plugin system to load the built-in plugin from outside the repository to be more independent.

## How I tested

1. Download [builtin-plugin-example.zip](https://github.com/reearth/reearth/files/12274290/builtin-plugin-example.zip)
2. Move to above file
3. `yarn && yarn dev`
4. Move to reearth/server
5. Set `REEARTH_EXT_PLUGIN=http://localhost:5001` in `.env` for server
6. Move to reearth/web
7. Set `REEARTH_WEB_UNSAFE_PLUGIN_URLS='["http://localhost:5001/Example.js"]'` in `.env` for web
8. `yarn start`
9. Open Beta project
10. Go to widget page
11. Add the widget named `Extension`


web/src/services/config/index.ts
```diff
@@ -80,9 +81,9 @@ export default async function loadConfig() {
     config.extensions = extensions;
   }
 
-  config.unsafeBuiltinPlugins = await loadUnsafeBuiltinPlugins();
-
   window.REEARTH_CONFIG = config;
+
+  config.unsafeBuiltinPlugins = await loadUnsafeBuiltinPlugins();
```
KaWaite, line 86: Why are you loading unsafeBuiltinPlugins after we assign config to window.REEARTH_CONFIG?

web/src/beta/lib/unsafeBuiltinPlugins/load.ts
```diff
@@ -0,0 +1,20 @@
+import { UnsafeBuiltinPlugin } from "./types";
+
+export const loadPlugins = async () => {
+  const urls = window.REEARTH_CONFIG?.unsafePluginUrls;
```
KaWaite, line 4: So basically you are processing and loading unsafePluginUrls twice, once in the config and then again here, right?
I'm guessing that is a mistake. Only one should be necessary, and if possible I'd prefer it be in the config. 🙏🏻

keiya01: You are right. I fixed it:pray:
[fix review](https://github.com/reearth/reearth/pull/625/commits/f72e6dfd7a69e50dad6d4af24828b42a4f926f16)

KaWaite (approved):

PR #626, opened 2023-08-07
chore(web): upgrade dependencies

airslice:
# Overview

This PR upgrades:

- `resium` `1.16.1` -> `1.17.1` (latest)
- `cesium` `1.105.2` -> `1.107.2`
Now both reearth and resium are based on cesium `1.107.2`.

Please pay attention to [this](https://community.cesium.com/t/cesiumjs-ready-promise-deprecation-api-changes/22469) change since `1.107`.

- `storybook` `7.0.18` -> `7.3.2` (latest)
- `vite` `4.3.2` -> `4.4.9` (latest)
Storybook should work now. 

- `@apollo/client` `3.7.14` -> `3.8.1` (latest)


web/.storybook/preview-head.html
```diff
@@ -0,0 +1,5 @@
+<style>
+  body{
+    overflow-y: auto !important;
```
keiya01, line 3: What is this?

web/src/beta/lib/core/engines/Cesium/Feature/Tileset/hooks.ts
```diff
@@ -431,13 +438,6 @@ export const useHooks = ({
         return;
       }
 
-      try {
```
keiya01, line 434: Maybe we need to test the clipping box as well.

web/.storybook/preview-head.html
airslice: This is for docs page, it's using GlobalStyles which contains body overflow hidden and making the page not scrollable. therefore i added this to fix it.

web/src/beta/lib/core/engines/Cesium/Feature/Tileset/hooks.ts
airslice: I've tested clipping box using view2.0 project, seems working well.

web/.storybook/preview-head.html
KaWaite: Where do we want it to scroll? The storybook UI has a navbar, action section (bottom), side panel, etc. I think it is on purpose to have the body not scrollable...

web/.storybook/preview.tsx
```diff
@@ -39,6 +40,9 @@ const preview: Preview = {
     layout: "fullscreen",
     controls: { expanded: true },
     actions: { argTypesRegex: "^on.*" },
+    docs: {
+      theme: themes.dark,
+    },
```
KaWaite, line 45: In `theme.ts` we create our own theme. Can you use that with docs too?

airslice: Updated.

web/package.json
```diff
@@ -74,6 +73,7 @@
         "eslint": "8.40.0",
         "eslint-config-reearth": "0.2.2",
         "eslint-plugin-playwright": "0.12.0",
+        "eslint-plugin-storybook": "^0.6.13",
```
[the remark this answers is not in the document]
airslice: Updated.

web/.storybook/preview-head.html
airslice: This is used for enable page scroll for docs page. I've updated the css so that it should not effect the story preview.

KaWaite (approved):

keiya01 (approved):
LGTM!

PR #628, opened 2023-08-10
chore(server): add missing fields for story settings

yk-eukarya:


server/pkg/storytelling/possision.go
```diff
@@ -0,0 +1,9 @@
+package storytelling
```
pyshx, line 1: I think there's a typo in the file name.

pyshx (approved):

PR #631, opened 2023-08-16
chore(web): TabsMenu for switching between tabs 

jashanbhullar:
# Overview
A right panel component with tabs which can switch between views

image

## What I've done
- Created the component 
- Created the story with some demo props

## What I haven't done
NA

## How I tested
Visually 

- Overall approach (as it's my first PR) 
- Do I need to write unit tests?

## Memo
NA


web/src/beta/components/TabMenu/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+    </Wrapper>
+  );
+};
+
+export default TabMenu;
+
+const Wrapper = styled.div`
+  width: 286px;
+  display: grid;
+  /* outline: solid red; */
```
KaWaite, line 49: Please remove any unneeded code

web/src/beta/components/TabMenu/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+  component: ReactNode;
+}
+
+type Props = {
+  title: string;
+  tabs: Record<string, TabObject>;
+};
+
+const TabMenu: FC<Props> = ({ tabs, title }) => {
+  const [selected, setSelected] = useState<string | null>(null);
```
KaWaite, line 18: Can you move this out of this component and instead have 2 props: 
```
selectedTab: string;
onSelectedTabChange: (tab: string) => void;
```
I think we will want to know which tab is selected outside of this component for knowing what to query/etc

web/src/beta/components/TabMenu/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { styled } from "@reearth/services/theme";
+
+interface TabObject {
+  icon: keyof typeof Icons;
+  component: ReactNode;
+}
+
+type Props = {
+  title: string;
+  tabs: Record<string, TabObject>;
```
KaWaite, line 14: If you make tabs just an array of objects, you can map from tabs itself instead of doing `Object.entries.....`

web/src/beta/components/TabMenu/index.tsx
KaWaite, line 1: Like I mentioned, the left and right panel we already have (which are resizeable), as well as the collapsible section component. 
This current implementation is ignoring those.
Please look at the current left/right panel implementation and consider how the tabbing can be applied there.
If you need help, let me know and we can pair pro 👍🏻

web/src/beta/components/TabMenu/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+};
+
+const TabMenu: FC<Props> = ({ tabs, title }) => {
+  const [selected, setSelected] = useState<string | null>(null);
+
+  return (
+    <Wrapper>
+      <Title>{title}</Title>
```
KaWaite, line 22: For text we have a Text component in the beta/components/ directory. Please use that

web/src/beta/components/TabMenu/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <Tabs>
+        {Object.entries(tabs).map(([tab, val]) => (
+          <Icon
+            key={tab}
+            onClick={() => setSelected(tab)}
+            icon={val.icon}
+            alt="icon"
+            size={20}
+            style={{
```
KaWaite, line 31: Please avoid inline styles as much as possible.

web/src/beta/components/TabMenu/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+  grid-column: 1/2;
+  background: black;
+  cursor: pointer;
+`;
+
+const MainArea = styled.div`
+  grid-column: 2/-1;
+  display: block;
+  padding: 6px;
+  background: #232226;
```
KaWaite, line 75: FYI, for colors use the theme. You can do so inside of a styled component like this:
```
const MainArea = styled.div`
  background: ${({ theme }) => theme.bg[1]};
`
```

web/src/beta/components/TabMenu/index.tsx
jashanbhullar: I wanted to use objects rather than array so it's easier to select the tabs with a <key, value> config. If you think arrays are better, I'll switch to that!

web/src/beta/components/TabMenu/index.tsx
jashanbhullar: The icon accepts a style component, that's why I was able to use it. I'll think of some other way if this is not correct.

web/src/beta/components/TabMenu/index.tsx
jashanbhullar: When selecting a tab it's also a better than using `index` or `find` to set the right tab especially if we are handling the `setting` outside this component

web/src/beta/components/TabMenu/index.tsx
jashanbhullar: Done!

web/src/beta/components/TabMenu/index.tsx
jashanbhullar: Done. 
Just for the `stories part` to dynamically update the tab when the user clicks on the icon, I have used `useArgs` , reference from [this blog](https://medium.com/urbandataanalytics/changing-args-values-state-in-storybook-without-usestate-hook-15cc096716f7), I hope this is the correct implementation as per our standards. I couldn't find a similar example that existed in the current codebase.

web/src/beta/components/TabMenu/index.tsx
jashanbhullar: So per my understanding, these can be wrapped in the ResizableComponent and we are good to go? is that correct?

web/src/beta/components/TabMenu/index.tsx
jashanbhullar: Done. Thanks :)

web/src/beta/components/TabMenu/index.tsx
jashanbhullar: How do i select which color to use? The hex codes in figma doesn't seem to be matching with theme.bg[x] for any color in the darktheme at least. I have added the values eyeballing close to how it should look like but IMO it can be done better. 
Please let me know if this is correct or need to be updated.

web/src/beta/components/TabMenu/index.tsx
KaWaite: I believe so. That, plus updating the css appropriately. Right now you are setting height and width with fixed values, but we want this to act like other content (refer to other content that is currently inside the left/right panels).

web/src/beta/components/TabMenu/index.tsx
KaWaite: Unless impossible, its nicer to have all custom styles inside a styled component, so just do `const StyledIcon = styled(Icon)`..styles..`

web/src/beta/components/TabMenu/index.tsx
KaWaite: The colors should match in figma. The theme was recently updated to mirror the theme in figma as much as possible.
If you notice any colors not looking the same as figma, talk with @nina992 and/or @lavalse (Red)

web/src/beta/components/TabMenu/index.tsx
KaWaite: Should be fine for now, though I think you can write your own args (not quite sure how this useArgs is working).
Screenshot 2023-08-17 at 16 01 58

web/src/beta/components/TabMenu/index.tsx
KaWaite: Thats fair. If this wasn't a prop field I'd say keep it as a record, but since its a prop, I feel like we should keep a similar syntax standard and up until now (and seen in many UI libraries, like Ant Design) we have done an array of objects that we iterate over.

web/src/beta/components/TabMenu/index.tsx
jashanbhullar: it's so that `onSelectedTabChange` function can work.

web/src/beta/components/TabMenu/index.tsx
jashanbhullar: Fair. Will make the change to Array!

jashanbhullar: This is done!

web/src/beta/components/TabMenu/index.tsx
jashanbhullar: Done!

web/src/beta/components/TabMenu/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+const TabMenu: FC<Props> = ({ tabs, title, selectedTab, onSelectedTabChange }) => {
+  const theme = useTheme();
+
+  const selectedTabItem = tabs.find(({ key }) => key === selectedTab);
+
+  return (
+    <Wrapper>
+      <Title>
+        <Text size="body">{title}</Text>
+      </Title>
```
KaWaite, line 30: We don't need this. Please look at the left/right panel implementations we already have to see how they are built up. 
Reference `src/beta/features/Editor/tabs/widgets/SidePanel/index.tsx` to see how it should be used (it'd be passed as the `children` prop)

web/src/beta/components/TabMenu/index.tsx
jashanbhullar: I have removed the hardcoded values from this!

web/src/beta/components/TabMenu/index.tsx
jashanbhullar: Changed the implementation. Thanks for this. 
Waiting for you to resolve the other comments on this!

web/src/beta/components/TabMenu/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+type Props = {
+  tabs: TabObject[];
+  selectedTab: string;
+  onSelectedTabChange: (tab: string) => void;
+};
+
+const TabMenu: FC<Props> = ({ tabs, selectedTab, onSelectedTabChange }) => {
+  const theme = useTheme();
+
+  const selectedTabItem = tabs.find(({ id }) => id === selectedTab);
```
KaWaite, line 24: useMemo

jashanbhullar: Done!

KaWaite (approved):

nina992 (approved):

PR #632, opened 2023-08-21
chore(server): add NLS layer support

pyshx:
# Overview
This PR adds support for NLS layers in the server and consider nls layer(aka newLayers in scene) as a seperate entity from the already existing layers.


server/go.mod
```diff
@@ -164,4 +164,4 @@ require (
 	moul.io/http2curl v1.0.1-0.20190925090545-5cd742060b0e // indirect
 )
 
-go 1.20
+go 1.19
```
yk-eukarya, line 167:
```suggestion
go 1.20
```

server/pkg/nlslayer/nlslayer.go
```diff
@@ -0,0 +1,146 @@
+package nlslayer
+
+import (
+	pl"github.com/reearth/reearth/server/pkg/layer"
+)
+
+type NLSLayer interface {
+	ID() ID
+	LayerType() string
```
yk-eukarya, line 9: It's more clear to have a `LayerType` type

server/pkg/nlslayer/nlslayer.go
```diff
@@ -0,0 +1,146 @@
...
+
+type layerBase struct {
+	id        		ID
+	layerType		string
+	scene     		SceneID
+	title			string
+	visible			bool
+	infobox			*pl.Infobox
+	tags      		*pl.TagList
+	creator 		string
```
yk-eukarya, line 66: Its better to keep the user id instead of his name, and then resolve it to the user info

server/pkg/nlslayer/nlslayer.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (l *layerBase) LayerType() string {
+	if l == nil {
+		return ""
+	}
+	return l.layerType
+}
+
+
+func (l *layerBase) Scene() SceneID {
```
yk-eukarya, line 88:
```suggestion
func (l *layerBase) Scene() SceneID {
	if l == nil {
		return ""
	}
```

server/pkg/nlslayer/nlslayer_list.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	for _, l := range layers {
+		if l == nil {
+			continue
+		}
+		l2 := *l
+		if l2 == nil {
+			continue
+		}
+		m[l2.ID()] = l
```
yk-eukarya, line 211: I did not get the idea here!

server/pkg/nlslayer/nlslayer_list.go
```diff
@@ -0,0 +1,296 @@
...
+	return list
+}
+
+func (m Map) Clone() Map {
+	if m == nil {
+		return Map{}
+	}
+	m2 := make(Map, len(m))
+	for k, v := range m {
+		m2[k] = v
```
yk-eukarya, line 233:
```suggestion
		m2[k] = v.Clone()
```

server/pkg/nlslayer/id_list.go
```diff
@@ -0,0 +1,223 @@
+// moved from github.com/reearth/reearth/server/pkg/layer
+package nlslayer
+
+import "github.com/reearth/reearth/server/pkg/id"
+
+type IDList struct {
```
yk-eukarya, line 6: Maybe its better to have a generic IDList type

server/pkg/nlslayer/data_time.go
```diff
@@ -0,0 +1,25 @@
...
+
+func (t *Time) SetUpdateClockOnLoad(updateClockOnLoad *bool) {
+	t.UpdateClockOnLoad = updateClockOnLoad
+}
+
+func (t *Time) Update(property *string, interval *int, updateClockOnLoad *bool) {
+	t.Property = property
+	t.Interval = interval
+	t.UpdateClockOnLoad = updateClockOnLoad
+}
```
yk-eukarya, line 25: what if `t` is nil?

pyshx: Are you suggest that I should add construction function like
```
func NewTime() *Time {
   return &Time{}
}
```

or I should add the checks at the beginning of each method? like
```
func (t *Time) SetProperty(property *string) {
   if t == nil { return }
   t.Property = property
}
```

server/pkg/nlslayer/nlslayer.go
pyshx: Are suggesting that creator name shouldn't be here but in another user info struct?

server/pkg/nlslayer/nlslayer_list.go
pyshx: 1. If the map (m) is nil, it initializes it to an empty map of the right type.
2. It iterates over the provided layers. For each layer:
  - If the layer pointer (l) is nil, it skips to the next iteration.
  - Dereferences the layer pointer to get the layer object (l2).
  - If l2 is nil (which means that the pointer points to a nil value), it skips to the next iteration.
  - Uses the ID() method of the layer (l2.ID()) to fetch its unique ID and adds the layer pointer to the map with that ID as the key.
3. Finally, it returns the modified (or new) map.

server/gql/newlayer.graphql
```diff
@@ -0,0 +1,149 @@
...
+    parentLayerId: ID!
+    layerType: String!
+    sceneID: ID!
+    dataType: String
+    dataUrl: URL
+    dataValue: Any
+    dataLayers: Any
+    dataJsonProperties: [String]
+    dataUpdateInterval: Int
+    dataParameters: Any
```
yk-eukarya, line 98:
```suggestion
    data: LayerDataInput
```

server/gql/newlayer.graphql
```diff
@@ -0,0 +1,149 @@
...
+    timeProperty: String
+    timeInterval: Int
+    timeUpdateClockOnLoad: Boolean
```
yk-eukarya, line 101:
```suggestion
    time: LayerTimeInput
```

server/pkg/nlslayer/group.go
```diff
@@ -0,0 +1,105 @@
+package nlslayer
+
+import (
+	pl"github.com/reearth/reearth/server/pkg/layer"
+)
+
+type NLSLayerGroup struct {
+	layerBase
+	children            *IDList
+	common				*LayerID
```
yk-eukarya, line 10: since `layerBase` contains `Id` why do we need this?

pyshx: This `LayerID` is for common layer https://github.com/reearth/reearth/blob/434b93671b469dbba5c40632656005a150307b42/server/gql/newlayer.graphql#L22

server/gql/scene.graphql
```diff
@@ -13,6 +13,8 @@ type Scene implements Node {
   team: Team
   property: Property
   rootLayer: LayerGroup
+  newLayers: [NLSLayer!]!
+  stylesList: [Any!]! # WIP
```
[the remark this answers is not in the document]
pyshx: This is to support Appearance collection that user can select out of.

server/pkg/nlslayer/nlslayer.go
pyshx: Can't do this for ID.

yk-eukarya (approved):

PR #650, opened 2023-08-29
chore(web): radio and radio group field components

nina992:
# Overview
Add Radio Box , Radio group to beta version of re-earth 
## What I've done
- Create Radio Box component 
- Create Radio Group component 
- Create Radio group field 

image
image
image
image
image


web/src/beta/components/RadioGroup/index.tsx
KaWaite, line 1: Can you move RadioBox in to the RadioGroup directory? Not sure why we would want to use it separately, so I think we should consider its only purpose is to make RadioGroup.
Also, I feel like this still should be considered a field. Even if it isn't a property field inside the layer adding modal, I think we can consider it a field. What do you think? If you agree, please move it in to the fields/ directory 🙏🏻

nina992: Done

pyshx (approved):

KaWaite (approved):

PR #651, opened 2023-08-30
feat(web): Toggle Field

jashanbhullar:
# Overview
Toggle Field implementation 

## What I've done
- basic toggle component
- Toggle field property

## What I haven't done
- NA

## How I tested
- stories

- onChange method
- styling
- Property field integration


nina992 (approved):

web/src/beta/features/Editor/tabs/publish/Nav/PublishModal/index.tsx
```diff
@@ -174,7 +174,7 @@ const PublishModal: React.FC<Props> = ({
           <HideableSection showOptions={showOptions}>
             <Wrapper>
               <Subtitle size="body">{t("Search engine indexing")}</Subtitle>
-              <ToggleButton checked={searchIndex} onChange={handleSearchIndexChange} />
+              <Toggle checked={searchIndex} onChange={handleSearchIndexChange} />
```
KaWaite, line 177: I think this toggle should be the field/Toggle. The design is as shown in the screenshot 🙏🏻

KaWaite: Of course don't set the description as `Description around`. That seems to be just boiler text since maybe @lavalse didn't have time to think what the descriptions should be in many places. We can add those later easily enough 👍🏻

web/src/beta/components/fields/ToggleField/index.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
```
KaWaite, line 1: Don't need React imports. Vite uses a plugin that injects them on build/runtime

web/src/beta/components/fields/ToggleField/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import Property from "@reearth/beta/components/fields";
+import Toggle from "@reearth/beta/components/Toggle";
+
+type ToggleProps = React.ComponentProps<typeof Toggle>;
```
KaWaite, line 6: Instead of this, can you just export Props from components/Toggle and import here from that component? 🙏🏻

web/src/beta/components/fields/PropertyFields/index.tsx
```diff
@@ -18,31 +19,45 @@ const PropertyFields: React.FC<Props> = ({ propertyId, item }) => {
...
-              value={(value as string) ?? ""}
-              description={sf.description}
-              onChange={handlePropertyValueUpdate(item.schemaGroup, propertyId, sf.id, sf.type)}
-            />
-          )
-        ) : (
-          <p key={sf.id}>{sf.name} field</p>
-        );
+
+        switch (sf.type) {
```
KaWaite, line 23: HmmMmmmMmmmMmmm thanks for giving this suggestion. I gave it some thought but still can't decide which is more readable....ternary operator is more condensed since we don't have to specify returns every time and also separate cases on individual lines...also, even with switch/case we would need to do some nesting, either way is not super easy to read.
Hmmm can you leave it as-is for now and we can come back to this later in a refactor or if it becomes hard to manage the more field components we add? 🙏🏻

web/src/beta/components/fields/ToggleField/index.tsx
jashanbhullar: That was what I initially did but then I thought rather than having multiple exports, this made more sense but I understand that that's more `simpler` and easy to understand.

web/src/beta/components/fields/ToggleField/index.tsx
jashanbhullar: Vite sounds nice.

web/src/beta/features/Editor/tabs/publish/Nav/PublishModal/index.tsx
jashanbhullar: image
It's in the publish page so it looks like this. Let me know if this design is changed, I kept it similar to as before.

web/src/beta/components/Toggle/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  );
+};
+
+export default Toggle;
+
+const Wrapper = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+`;
```
KaWaite, line 32: This isn't needed anymore so remove it 🙏🏻

web/src/beta/features/Editor/tabs/publish/Nav/PublishModal/index.tsx
```diff
@@ -173,8 +173,13 @@ const PublishModal: React.FC<Props> = ({
           </OptionsToggle>
           <HideableSection showOptions={showOptions}>
             <Wrapper>
-              <Subtitle size="body">{t("Search engine indexing")}</Subtitle>
-              <ToggleButton checked={searchIndex} onChange={handleSearchIndexChange} />
+              <ToggleField
+                name={t("Search engine indexing")}
+                // TODO: Add japanese translation for the text below
```
KaWaite, line 178: Don't need this comment

KaWaite (approved):

web/src/beta/components/Toggle/index.tsx
jashanbhullar: Oh yea, sorry :sweat:

PR #652, opened 2023-08-31
chore(web): Slider Field

jashanbhullar:
# Overview
Sliders field implementation

## What I've done
- Classic code migrated 
- Colors updated 
- Disable state fixed
- Added property field
- Tooltip added

## What I haven't done
- NA

## How I tested
- Stories

## Which point do I want you to review particularly
- UI implementation 
- Property field

## Memo
- NA


web/src/beta/components/Slider/index.tsx
```diff
@@ -1,52 +1,55 @@
...
-type Props = {
+export type Props = {
   min: number;
   max: number;
-  frame?: boolean;
 } & Omit<ComponentProps<typeof RCSlider>, "defaultValue">;
 
-const Slider: React.FC<Props> = ({ frame = false, ...props }) => (
-  <Wrapper frame={frame}>
+// TODO: Show tooltip on hover and drag
```
jashanbhullar, line 13: To implement tooltip either I can build a small `custom component` or use `rc-tooltip` from the example at https://github.com/react-component/slider/blob/master/docs/examples/components/TooltipSlider.tsx ? 

Let me know which one works. I don't want to add a dependency for a small functionality but if it's going to be used over and over again it makes sense

web/src/beta/components/fields/PropertyFields/index.tsx
```diff
@@ -19,6 +20,7 @@ const PropertyFields: React.FC<Props> = ({ propertyId, item }) => {
       {item?.schemaFields.map(sf => {
         const isList = item && "items" in item;
         const value = !isList ? item.fields.find(f => f.id === sf.id)?.value : sf.defaultValue;
+        const { min, max } = sf;
```
jashanbhullar, line 23: Is this okay or should there be defaults for `0` and `100` for the min-max values?

web/src/beta/components/Slider/index.tsx
KaWaite: Go with the example you reference. rc-tooltip is a dep of rc-slider so you don't need to add any new dep 👍🏻

web/src/beta/components/fields/PropertyFields/index.tsx
KaWaite: I don't think we should set defaults because we don't know what the slider field will be pointing to. 
min/max values should be set for things like the slider. If they aren't set, we should add them to the manifest in the backend 👍🏻 
Let me know if you are getting null/undefined for any values related to the slider 🙏🏻

jashanbhullar: Umm..I don't know where they are being used so I don't know how to check!

web/src/beta/components/Slider/index.tsx
jashanbhullar: Thanks! :+1:

web/src/beta/components/fields/PropertyFields/index.tsx
nina992: I think we used min/max once with spacing so we can define the min and max padding for an item (like image block ) for example

web/src/beta/components/fields/SliderField/index.tsx
```diff
@@ -0,0 +1,17 @@
+import Property from "@reearth/beta/components/fields";
+import Slider, { Props as ToggleProps } from "@reearth/beta/components/Slider";
+
+export type Props = {
+  name?: string;
+  description?: string;
+} & ToggleProps;
```
nina992, line 7: why don't  we use a general props like adding type file and import the shared props from it , instead of mentioned another field props

web/src/beta/components/fields/PropertyFields/index.tsx
```diff
@@ -61,6 +63,16 @@ const PropertyFields: React.FC<Props> = ({ propertyId, item }) => {
             description={sf.description}
             onChange={handlePropertyValueUpdate(item.schemaGroup, propertyId, sf.id, sf.type)}
           />
+        ) : sf.type == "number" ? (
```
KaWaite, line 66: sf.type === "number" ? sf.ui === "slider" ?

web/src/beta/components/fields/PropertyFields/index.tsx
KaWaite: Sorry can you remove this and just use min/max like other sf values (ie sf.min, sf.max)

web/src/beta/components/fields/SliderField/index.tsx
jashanbhullar: You mean for the name and description?

nina992 (approved):

web/src/beta/components/fields/PropertyFields/index.tsx
KaWaite: Notice we want three equals, not two. You are doing two equals

PR #680, opened 2023-09-11
refactor: use radio group in Data source manage

nina992:
# Overview
Refactor data source manager to use Radio group component instead the classic one 
## What I've done
 
image
image


web/src/beta/components/RadioGroup/RadioBox/index.tsx
```diff
@@ -23,7 +24,7 @@ const RadioBox: React.FC<Props> = ({ inactive, selected, value, onClick }: Props
       <RadioButton selected={isChecked} inactive={inactive}>
         {isChecked && <Checkmark selected={isChecked} inactive={inactive} />}
       </RadioButton>
-      <RadioText>{value}</RadioText>
+      <RadioText>{label ?? value}</RadioText>
```
KaWaite, line 27: What we show for the radio button should never be the value I think. The label should be internationalized (and possibly capitalized depending on the language). Saved values will most definitely be keys, so can you update this component in that way?
So
- value -> key
- pass key in to onClick()
- use label only for RadioText
- Make key and label non-optional

web/src/beta/components/RadioGroup/RadioBox/index.tsx
KaWaite, line 1: Your onClick in RadioBox accepts a value prop, but then in RadioGroup you do `() => handleRadioChange(option.value)`. Either approach is okay, but please stick to one (so update the typing on RadioBox, or just pass handleRadioChange without the () => so that RaadioBox uses it as it should

web/src/beta/features/Editor/DataSourceManager/Asset/index.tsx
```diff
@@ -1,9 +1,10 @@
 import React from "react";
 
 import Button from "@reearth/beta/components/Button";
+import RadioGroup from "@reearth/beta/components/RadioGroup";
 import Toggle from "@reearth/beta/components/Toggle";
 import generateRandomString from "@reearth/beta/utils/generate-random-string";
-import RadioButton from "@reearth/classic/components/atoms/RadioButton";
+//import RadioButton from "@reearth/classic/components/atoms/RadioButton";
```
KaWaite, line 7: Delete

web/src/beta/features/Editor/DataSourceManager/Asset/index.tsx
```diff
@@ -61,22 +61,14 @@ const Asset: React.FC<DataProps> = ({ sceneId, onSubmit, onClose }) => {
...
-                value="value"
-                checked={sourceType == "value"}
-                handleChange={c => c && setSourceType("value")}
-              />
-              From Value
-            </RadioButtonLabel>
+            <RadioGroup
+              options={[
+                { label: "From URL", value: "url", selected: true },
+                { label: "From Value", value: "value", selected: false },
```
KaWaite, line 67: i18n please

web/src/beta/components/RadioGroup/RadioBox/index.tsx
nina992: sorry but I am a little confused about the key and value , when you said value=> key , you mean only changing the name right ? , aaswe need separate value whatever its name and sometimes it isdifferent that the label for example the case of data source manager

nina992: About passing the key or value to onclick isn't the same as it is now as we have value param and we already passed it to onclick

web/src/beta/components/RadioGroup/RadioBox/index.tsx
nina992: done

web/src/beta/features/Editor/DataSourceManager/Asset/index.tsx
nina992: done

web/src/beta/features/Editor/DataSourceManager/Asset/index.tsx
nina992: done

web/src/beta/components/RadioGroup/RadioBox/index.tsx
KaWaite: yeah value -> key means name change. Let me know if anything else doesn't make sense

web/src/beta/components/RadioGroup/RadioBox/index.tsx
```diff
@@ -7,23 +7,31 @@ export type Props = {
...
   const [isChecked, setIsChecked] = useState(selected ?? false);
 
-  const handleRadioClick = useCallback(() => {
-    setIsChecked(!isChecked);
-    if (onClick) onClick(value);
-  }, [isChecked, onClick, value]);
+  const handleRadioClick = useCallback(
+    (value: string) => {
+      setIsChecked(!isChecked);
+      if (onClick) onClick(value);
```
KaWaite, line 19: Just doing `onClick?.(value)` is enough. If there is an onClick, it'll be used, and if not it of course won't be used 👍🏻

web/src/beta/components/RadioGroup/RadioBox/index.tsx
```diff
@@ -7,23 +7,31 @@ export type Props = {
   selected?: boolean;
   value: string;
   onClick?: (value: string) => void;
+  label?: string;
 };
 
-const RadioBox: React.FC<Props> = ({ inactive, selected, value, onClick }: Props) => {
+const RadioBox: React.FC<Props> = ({ inactive, selected, value, label, onClick }: Props) => {
```
KaWaite, line 13: the `inactive` prop doesn't seem to be used in RadioGroup. Please remove the logic around it

web/src/beta/components/RadioGroup/RadioBox/index.tsx
```diff
@@ -7,23 +7,31 @@ export type Props = {
...
   const [isChecked, setIsChecked] = useState(selected ?? false);
```
KaWaite, line 14: you can just do `useState(!!selected)`

web/src/beta/components/RadioGroup/RadioBox/index.tsx
```diff
@@ -7,23 +7,31 @@ export type Props = {
...
   return (
     <Radio selected={isChecked} inactive={inactive}>
-      <RadioInput type="radio" value={inactive ? undefined : value} onClick={handleRadioClick} />
+      <RadioInput
+        type="radio"
+        value={inactive ? undefined : value}
+        onClick={() => handleRadioClick(value)}
+      />
       <RadioButton selected={isChecked} inactive={inactive}>
         {isChecked && <Checkmark selected={isChecked} inactive={inactive} />}
```
KaWaite, line 32: Checkmark isn't a checkmark. Please rename.

web/src/beta/components/RadioGroup/index.tsx
```diff
@@ -4,7 +4,7 @@ import RadioBox from "@reearth/beta/components/RadioGroup/RadioBox";
 import { styled } from "@reearth/services/theme";
 
 export type Option = {
```
KaWaite, line 6: Can we remove selected from Option and pass it as a single string prop whose value is the selected option's value?
If we do that, we can avoid the `isAtLeastOneSelected` stuff. 
We can then also replace currentOptions with currently selected state that does `useState(selected ?? options[0].value)`.

web/src/beta/components/RadioGroup/RadioBox/index.tsx
nina992: ah, but actually I don't thing we should do that first of all is a reserved keyword and doesn't make sense to be prop in a component , for example radio box is used like this ` <RadioGroupContainer layout={layout}>
      {currentOptions.map((option, index) => (
        <RadioBox
          key={`${option.label}-${key}-${index}`}
          value={option.value}
          selected={option.selected}
          label={option.label}
          onClick={() => handleRadioChange(option.value)}
        />
      ))}
    </RadioGroupContainer>` potential confusion

nina992: anyway I renamed it to be keyValue

web/src/beta/components/RadioGroup/RadioBox/index.tsx
nina992: done

web/src/beta/components/RadioGroup/RadioBox/index.tsx
nina992: done

web/src/beta/components/RadioGroup/RadioBox/index.tsx
nina992: done

web/src/beta/components/RadioGroup/RadioBox/index.tsx
nina992: done

web/src/beta/components/RadioGroup/index.tsx
```diff
@@ -16,38 +15,43 @@ export type RadioGroupProps = {
 };
 
 const RadioGroup: React.FC<RadioGroupProps> = ({ options, layout, onChange }) => {
-  const [currentOptions, updateOptions] = useState<Option[]>(options);
   const [key, setKey] = useState(0);
```
KaWaite, line 18: Sorry I just noticed this key,setKey thing. What is this? It doesn't seem to do anything meaningful. Can you explain it or delete it? 🙏🏻

nina992: The use of key and setKey in this code helps ensure that the RadioBox components are properly updated and re-rendered when the selected value changes, It ensures that each RadioBox maintains its own state and behaves correctly in response to changes in the parent component's state. if I removed it the select doesn't work right and the component doesn't re-render

web/src/beta/components/RadioGroup/RadioBox/index.tsx
```diff
@@ -5,46 +5,45 @@ import { fonts, styled } from "@reearth/services/theme";
...
-  value: string;
+  keyValue: string;
   onClick?: (value: string) => void;
+  label?: string;
 };
 
-const RadioBox: React.FC<Props> = ({ inactive, selected, value, onClick }: Props) => {
-  const [isChecked, setIsChecked] = useState(selected ?? false);
+const RadioBox: React.FC<Props> = ({ selected, keyValue, label, onClick }: Props) => {
+  const [isChecked, setIsChecked] = useState(!!selected);
```
KaWaite, line 14: Sorry can you update the name here too since we aren't checking it, we are selecting. So `isSelected`

web/src/beta/components/RadioGroup/index.tsx
```diff
@@ -4,50 +4,49 @@ import RadioBox from "@reearth/beta/components/RadioGroup/RadioBox";
...
+      {options.map((option, index) => (
         <RadioBox
-          key={`${option.key}-${key}`}
-          value={option.value}
-          selected={option.selected}
-          onClick={() => handleRadioChange(option.value)}
+          key={`${option.keyValue}-${key}-${index}`}
+          keyValue={option.keyValue}
+          selected={option.keyValue === selectedValue}
+          label={option.label || option.keyValue}
```
KaWaite, line 38: We don't want label to ever be keyValue. The keyValue will be stuff like "url", "csv", "default_tile", etc. We don't want to display that. We will want label to be required 🙏🏻

web/src/beta/components/RadioGroup/index.tsx
KaWaite: Are you 100% sure? Did you try to remove `key, setKey` before or after you removed `currentOptions`? 
Because I feel like while using `currentOptions`, what you say makes sense since we don't update `currentOptions` in a useEffect after `options` changes, and we also had the selected value in the `currentOptions`.
But now that `selectedValue` is separate, and we use `options` directly instead of `currentOptions`, I feel like this `key` isn't needed anymore....

web/src/beta/components/RadioGroup/index.tsx
```diff
@@ -4,50 +4,49 @@ import RadioBox from "@reearth/beta/components/RadioGroup/RadioBox";
...
 export type RadioGroupProps = {
   options: Option[];
   layout?: "vertical" | "horizontal";
+  selectedValue?: string;
   onChange?: (value: string) => void;
 };
 
-const RadioGroup: React.FC<RadioGroupProps> = ({ options, layout, onChange }) => {
-  const [currentOptions, updateOptions] = useState<Option[]>(options);
+const RadioGroup: React.FC<RadioGroupProps> = ({ options, layout, selectedValue, onChange }) => {
```
KaWaite, line 18: `layout = "horizontal"`

web/src/beta/features/Editor/DataSourceManager/Asset/index.tsx
```diff
@@ -78,22 +85,12 @@ const Asset: React.FC<DataProps> = ({ sceneId, onSubmit, onClose }) => {
...
-              <RadioButton
-                value="value"
-                checked={sourceType == "value"}
-                handleChange={c => c && setSourceType("value")}
-              />
-              {t("From Value")}
-            </RadioButtonLabel>
+            <RadioGroup
+              options={DataSourceOptions}
+              layout="horizontal"
```
KaWaite, line 90: If you do the above recommendation, then remove layout="horizontal" here

web/src/beta/features/Editor/DataSourceManager/DelimitedText/index.tsx
```diff
@@ -61,14 +63,11 @@ const DelimitedText: React.FC<DataProps> = ({ sceneId, onSubmit, onClose }) => {
...
-              <RadioButton
-                value="url"
-                checked={sourceType == "url"}
-                handleChange={c => c && setSourceType("url")}
-              />
-              From URL
-            </RadioButtonLabel>
+            <RadioGroup
+              options={[{ label: t("From URL"), keyValue: "url" }]}
+              layout="horizontal"
```
KaWaite, line 68: If you do the above recommendation, then remove layout="horizontal" here

web/src/beta/components/RadioGroup/index.tsx
nina992: yes I tried to remove it and all the select action ruined , actually I  face this issue during implement the radio group and this was  the only way to fix it

web/src/beta/components/RadioGroup/index.tsx
nina992: sorry sorry I forget to remove it ,sorry

web/src/beta/components/RadioGroup/index.tsx
nina992: done

web/src/beta/features/Editor/DataSourceManager/Asset/index.tsx
nina992: done

web/src/beta/features/Editor/DataSourceManager/DelimitedText/index.tsx
nina992: done

web/src/beta/components/RadioGroup/RadioBox/index.tsx
nina992: done

web/src/beta/components/RadioGroup/RadioBox/index.tsx
KaWaite: Please also remove it from Props type 🙏🏻

web/src/beta/components/RadioGroup/index.tsx
KaWaite: I found a solution: https://stackoverflow.com/questions/60271440/react-map-not-re-rendering

pyshx (approved):

web/src/beta/components/RadioGroup/RadioBox/index.tsx
nina992: done

web/src/beta/components/RadioGroup/index.tsx
nina992: done - since key value is mandatory

KaWaite (approved):

PR #689, opened 2023-09-14
feat(web): story drag and drop

mkumbobeaty:
# Overview

## What I've done
- drag and drop story 
- on scroll the story


web/src/beta/components/DragAndDropList/Item.tsx
```diff
@@ -91,7 +94,16 @@ const Item: FC<Props> = ({
...
+
+  // eslint-disable-next-line no-prototype-builtins
+  return item.hasOwnProperty("extensionId") ? (
+    <SItem ref={preview} data-handler-id={handlerId} isDragging={isDragging}>
+      <DndHandle ref={ref}>
+        <Icon icon="dndHandle" size={16} />
+      </DndHandle>
+      {children}
+    </SItem>
+  ) : (
```
KaWaite, line 106: This is:
1) making it so the handle Icon is ALWAYS showing, which isn't good.
2) can only drop successfully if I drag overtop of another item's handle.

Can you try to fix these two issues? 🙏🏻

web/src/beta/components/DragAndDropList/Item.tsx
```diff
@@ -15,6 +16,7 @@ type Props = {
   itemGroupKey: string;
   id: string;
   index: number;
+  item?: any;
```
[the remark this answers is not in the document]
KaWaite: Ah sorry, I guess you are only checking if the item has an extensionId, right? That seems like a lot of work to know if it is a StoryBlock or not. 
In actuality though, what you want to know is whether to make the WHOLE child the handle, or just the single element (in the StoryBlock's case, the dnd handle Icon, right?), right?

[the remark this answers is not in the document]
mkumbobeaty: If I won't check if is story telling the logic implemented here uses the ref which is used to drag and drop, I changed some of codes but I think we need to discuss so that I can explain why I did that

web/src/beta/components/DragAndDropList/Item.tsx
mkumbobeaty: fixed

web/src/beta/lib/core/StoryPanel/Page/index.tsx
```diff
@@ -47,82 +71,137 @@ const StoryPage: React.FC<Props> = ({
   onBlockDelete,
   onBlockSelect,
   onPropertyUpdate,
+  handleMoveBlock,
```
KaWaite, line 74: onBlockMove

web/src/beta/lib/core/StoryPanel/Page/index.tsx
```diff
@@ -47,82 +71,137 @@ const StoryPage: React.FC<Props> = ({
...
-            <StoryBlock
-              block={b}
-              isSelected={selectedStoryBlockId === b.id}
-              isEditable={isEditable}
-              onClick={() => onBlockSelect?.(b.id)}
-              onClickAway={onBlockSelect}
-              onChange={onPropertyUpdate}
-              onRemove={onBlockDelete}
-            />
+    <DndProvider>
```
KaWaite, line 115: DndProvider is already wrapping the Editor and Published components, so you shouldn't need to wrap here.

web/src/beta/lib/core/StoryPanel/Page/index.tsx
```diff
@@ -32,8 +45,19 @@ type Props = {
...
 
+const StoryPageContext = createContext({
+  isHovered: false,
+  handleOnMouseOut: () => {},
+  handleOnMouseOver: (e: MouseEvent<HTMLDivElement>) => {
+    e;
+  },
+});
+
+export const useStoryPageContext = () => useContext(StoryPageContext);
```
KaWaite, line 59: Can you give any big reasons you are using context here?
For one, I feel like it adds complexity since most things we need downstream we are passing as props (even if they need to go far).
But also, I think because you define and use hover state here at the page level, if I hover over one block, I get ALL handles showing in that page.

web/src/beta/lib/core/StoryPanel/Page/index.tsx
mkumbobeaty: @KaWaite  handleMoveBlock function is used to update MoveStoryBlock mutation but I think this one (onBlockMove) has other functionalities.

web/src/beta/lib/core/StoryPanel/Page/index.tsx
mkumbobeaty: @KaWaite I removed this

web/src/beta/lib/core/StoryPanel/Page/index.tsx
KaWaite: Sorry don't quite understand what you mean.
For me, I was just pointing out the naming convention is wrong. For component props, functions should be named `onNounVerb`, whereas locally defined functions should be `handleNounVerb`.
This helps with readability and is a React standard 🙏🏻

web/src/beta/components/DragAndDropList/Item.tsx
```diff
@@ -15,22 +15,30 @@ type Props = {
...
+  shouldUseCustomHandler?: boolean;
   onItemMove: (dragIndex: number, hoverIndex: number) => void;
   onItemDropOnItem: (dropIndex: number) => void;
   onItemDropOutside: () => void;
   children: ReactNode;
 };
 
+const ItemContext = createContext<React.RefObject<HTMLDivElement> | null>(null);
+
+export const useItemContext = () => useContext(ItemContext);
```
KaWaite, line 27: Sorry, similar to the other context, if there isn't a big reason (other than limit prop drilling) for this context can you just pass down as a prop. Later we might want to strategically use context in a refactor to cleanup/simplify the code.

web/src/beta/lib/core/StoryPanel/Page/index.tsx
```diff
@@ -97,36 +105,55 @@ const StoryPage: React.FC<Props> = ({
...
-                installableStoryBlocks={installableStoryBlocks}
-                onBlockOpen={() => handleBlockOpen(idx)}
-                onBlockAdd={onBlockCreate?.(idx + 1)}
-              />
-            )}
-          </Fragment>
-        ))}
+        {storyBlocks && storyBlocks.length > 0 && (
+          <DragAndDropList
+            uniqueKey="LeftPanelPages"
```
KaWaite, line 116: uniqueKey should be...unique :P lol should be maybe "storyPanel" or something

web/src/beta/lib/core/StoryPanel/Page/index.tsx
mkumbobeaty: @KaWaite there is already onBlockMove function so it will be a duplicate may be I can make it onBockDnd or onBlockMoved

web/src/beta/components/DragAndDropList/Item.tsx
mkumbobeaty: @KaWaite the reason am using context here is different from prev one I can explain to you tomorrow on call

KaWaite (approved):

nina992 (approved):

keiya01 (approved):

PR #693, opened 2023-09-18
chore(server): add extensionId and pluginId in block json of published story

pyshx:
# Overview
This PR adds pluginId and extensionId fields in the block JSON created while publishing story.


yk-eukarya (approved):
LGTM

PR #694, opened 2023-09-18
feat(web): Add layers using wms and mvt

mkumbobeaty:
# Overview

## What I've done
- Implemented add layer using mvt format
- implemented add layer using wms format


web/src/beta/features/Editor/DataSourceManager/index.tsx
```diff
@@ -37,6 +39,16 @@ const DataSourceManager: React.FC<DataProps> = ({ sceneId, onClose, onSubmit })
           id: "threeDTiles",
           label: "3D Tiles",
         },
+        {
+          content: <WmsTiles sceneId={sceneId} onSubmit={onSubmit} onClose={onClose} />,
+          id: "wms",
+          label: "WMS",
```
KaWaite, line 45: Can you wrap all the labels with i18n? 🙏🏻 Most will be the same in Japanese, but some will be different.

web/src/beta/features/Editor/DataSourceManager/WmsTiles/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+    setLayers(updatedLayers);
+  };
+
+  const handleSubmit = () => {
+    onSubmit({
+      layerType: "simple",
+      sceneId,
+      title: generateRandomString(5),
+      visible: true,
+      config: {
```
KaWaite, line 43: Can you and @pyshx discuss where might a good place be to write and export all format type's config objects? 
Even though I am not sure we need to use them in multiple places, I still feel like having them all in one place would be good 🙏🏻

web/src/beta/features/Editor/DataSourceManager/WmsTiles/index.tsx
```diff
@@ -0,0 +1,120 @@
+import { FC, useState } from "react";
+
+import Button from "@reearth/beta/components/Button";
+import generateRandomString from "@reearth/beta/utils/generate-random-string";
+
+import { DataProps } from "..";
+import {
+  AddLayerWrapper,
+  AssetWrapper,
+  ColJustiftBetween,
```
KaWaite, line 10: Typo: ColJustifyBetween

web/src/beta/features/Editor/DataSourceManager/WmsTiles/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+        },
+      },
+    });
+    onClose();
+  };
+
+  return (
+    <ColJustiftBetween>
+      <AssetWrapper>
+        <InputGroup label="Resource URL" description="URL of the data source you want to add.">
```
KaWaite, line 69: Can you wrap all user-facing text in i18n? 🙏🏻

pyshx (approved):
LGTM!

web/src/beta/features/Editor/DataSourceManager/hooks.ts
```diff
@@ -0,0 +1,46 @@
+import { useState, KeyboardEvent } from "react";
+
+export default () => {
+  const [urlValue, setUrlValue] = useState("");
+  const [layerValue, setLayerValue] = useState("");
+  const [layers, setLayers] = useState<string[]>([]);
+  const [layerInput, setLayerInput] = useState(false);
+
+  const handleonAddLayer = () => {
```
KaWaite, line 9: handleLayerAdd

web/src/beta/features/Editor/DataSourceManager/hooks.ts
KaWaite, line 1: Where is this being used?

mkumbobeaty: @KaWaite I am using this to because both WmsTiles and vectorTiles uses the same logic's of adding layers so I thought instead of write them in every component which is like duplicate, I better to have them in this file

PR #696, opened 2023-09-19
chore(web): Select field - Minor fixes

jashanbhullar:
# Overview
- Select field minor fixes

## What I've done
- added z-index 
- added reverting arrow icon 
- fixed height 
- added placeholder prop

## What I haven't done
NA

## How I tested
- Stories 

NA

## Memo
NA

### DIFF
--- a/web/src/beta/components/fields/SelectField/index.tsx
+++ b/web/src/beta/components/fields/SelectField/index.tsx
@@ -31,7 +31,6 @@ const SelectField: React.FC<Props> = ({
   disabled = false,
   name,
   description,
-  placeholder,
 }) => {
   const t = useT();
 
@@ -57,7 +56,7 @@ const SelectField: React.FC<Props> = ({
         <Popover.Trigger asChild>
           <InputWrapper disabled={disabled} onClick={handlePopOver}>
             <Select selected={selected ? true : false} open={open}>
-              {selected ? selected.label : placeholder ? placeholder : t("Please choose an option")}
+              {selected ? selected.label : t("Please choose an option")}
             </Select>
             <ArrowIcon icon="arrowDown" open={open} size={12} />
           </InputWrapper>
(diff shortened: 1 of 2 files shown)


web/src/beta/components/fields/SelectField/index.tsx
```diff
@@ -55,9 +57,9 @@ const SelectField: React.FC<Props> = ({
         <Popover.Trigger asChild>
           <InputWrapper disabled={disabled} onClick={handlePopOver}>
             <Select selected={selected ? true : false} open={open}>
-              {selected ? selected.label : t("Please choose an option")}
+              {selected ? selected.label : placeholder ? placeholder : t("Please choose an option")}
```
KaWaite, line 60:
```suggestion
              {selected ? selected.label : placeholder ?? t("Please choose an option")}
```

web/src/beta/components/fields/SelectField/index.tsx
```diff
@@ -86,8 +88,14 @@ const InputWrapper = styled.div<{ disabled: boolean }>`
...
   display: flex;
   padding: 4px 8px;
+  /* The width + placement of the arrow icon */
+  padding-right: 22px;
   border-radius: 4px;
   width: 100%;
+  height: 28px;
+  line-height: 2.33;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
KaWaite, line 98: Something in here is making the height much too big.

Also, I am not sure overflow hidden is a good solution. Does the design show a placeholder and/or do we want a placeholder for anywhere that the select is being used? I feel like we want to ALWAYS have a value, which can be a fallback/default value rather than a placeholder... no?

nina992 (approved):

web/src/beta/components/fields/SelectField/index.tsx
jashanbhullar: removed the placeholder

web/src/beta/components/fields/SelectField/index.tsx
jashanbhullar: Changed. it's working as expected now!

KaWaite (approved):

PR #703, opened 2023-09-25
chore(web): add line-height option to story text block

airslice:
# Overview

This PR adds line-height option (1.0-2.0) to story text block.
This PR also fixes a bug which leads to crash sometimes.


web/src/beta/lib/lexical/RichTextEditor/plugins/ToolbarPlugin.tsx
```diff
@@ -90,6 +90,20 @@ const FONT_SIZE_OPTIONS: [string, string][] = [
   ["20px", "20px"],
 ];
 
+const LINE_HEIGHT_OPTIONS: [string, string][] = [
```
KaWaite, line 93: Don't see why you don't just do `["1.0", "1.1", etc, etc]`?
Is it just in case we sometime want to display some different value to the user? or..?

KaWaite (approved):

web/src/beta/lib/lexical/RichTextEditor/plugins/ToolbarPlugin.tsx
airslice: I feel the same...As you know this plugin is basically referenced from the official demo and i guess the origin author did this for some reason so i kept the same.

PR #706, opened 2023-09-26
chore(web): update layer title default value in beta and update add layerName behaviour for mvt & wms

pyshx:
# Overview
This PR makes refactor around the layer title default value, the showing behaviour of layer title in the left panel and adds gql refactors coming from backend. 

It also fixes the add layer behaviour with `mvt` and `wms` data type. where if layerName is provided on the input field and on click submit it should be added to the array of layer name.


web/src/beta/features/Editor/tabs/map/LeftPanel/Layers/LayerItem.tsx
```diff
@@ -90,10 +91,19 @@ const LayerItem = ({
...
       )}
     </ListItem>
   );
 };
 
+const TitleContainer = styled.div`
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  max-width: 200px;
```
KaWaite, line 106: Since we can resize the left/right panels, we shouldnt set a max-width. It should grow to show more of the hidden layer name as the panel grows.
Also, the `...` isn't showing. The name just gets hard cutoff

pyshx: Fixed.

web/src/beta/features/Editor/DataSourceManager/utils.tsx
```diff
@@ -112,3 +113,23 @@ export const AddLayerWrapper = styled.div`
...
+      const fileName = lastPart.split(".")[0];
+      return fileName;
+    } catch (error) {
+      console.error("Invalid URL", error);
+    }
+  }
+
+  if (sourceType === "url") {
+    console.warn("SourceType is 'url' but either URL is invalid or not provided.");
+  }
```
KaWaite, line 132: Do we really need this and the sourceType param? If the try/catch gets an error we know something is wrong. Both the catch's error and this warning are goign to go off without much difference in meaning, right?

KaWaite (approved):

web/src/beta/features/Editor/DataSourceManager/utils.tsx
pyshx: Fair point, I'll remove it.

PR #714, opened 2023-09-28
chore(web): story page update

mkumbobeaty:
# Overview

## What I've done

- hide/show state shown in check-boxes
- on checkbox click update page with/without layerId
- update visualizer to show only certain layers depending on page selected


web/src/beta/features/Editor/Visualizer/index.tsx
```diff
@@ -20,7 +22,8 @@ export type Props = {
   // storytelling
   showStoryPanel?: boolean;
   selectedStory?: Story;
-  currentPageId?: string;
+  currentPage?: Page;
+  tab?: Tab;
```
KaWaite, line 26: We don't need tab added. We can know if its a story based on whether `showStoryPanel` is true or not 👍🏻

web/src/beta/features/Editor/Visualizer/index.tsx
```diff
@@ -89,6 +93,21 @@ const Visualizer: React.FC<Props> = ({
...
+      const results =
+        currentPage?.layersIds?.length && (tab === "story" || showStoryPanel)
+          ? filteredLayers
+          : layers || [];
+      setLayersData(results);
+    };
+
+    handleDisplayLayer();
+  }, [currentPage, layers, tab, showStoryPanel]);
+
```
KaWaite, line 110: Can you move this in to the hooks file?

web/src/beta/features/Editor/Settings/index.tsx
```diff
@@ -27,22 +30,49 @@ const Settings: React.FC<Props> = ({
...
+  }, [selectedPage, layers]);
+
+  const pageId = selectedPage?.id;
+  const handleLayerCheck = (layerId: string) => {
+    setLayerChecked(prev => {
+      const updatedLayers = prev.includes(layerId)
+        ? prev.filter(id => id !== layerId)
+        : [...prev, layerId];
+
+      pageId && onPageUpdate?.(pageId, updatedLayers);
```
KaWaite, line 54: To avoid the UI becoming incorrect we should:
- `if (!pageId) return;` at the top of handleLayerCheck
- return prev if onPageUpdate fails <- this is maybe ignorable if the performance becomes really bad... in the future I want to implement optimistic mutations, but until then the feedback of user's click might be too slow....

web/src/beta/features/Editor/Visualizer/hooks.ts
```diff
@@ -240,6 +245,19 @@ export default ({
...
   }, [sceneProperty]);
 
+  const [layersData, setLayersData] = useState(layers);
+
+  useEffect(() => {
+    const handleDisplayLayer = () => {
+      const filteredLayers = layers?.filter(layer => currentPage?.layersIds?.includes(layer.id));
+      const layerPages = (currentPage?.layersIds?.length ?? 0) > 0 ? filteredLayers : [];
+      const results = showStoryPanel ? layerPages : layers;
+      setLayersData(results);
```
KaWaite, line 255: I don't think we want to filter layers ALWAYS when we might not need to.
Can you check 
```
if (!!showStoryPanel) { 
   setLayersData(layers);
   return;
   };
``` 
at the start of the useEffect so we can avoid the processing 🙏🏻

web/src/beta/features/Editor/Settings/hooks.ts
```diff
@@ -0,0 +1,67 @@
...
+import { NLSLayer } from "@reearth/services/api/layersApi/utils";
+import { Page } from "@reearth/services/api/storytellingApi/utils";
+
+type SettingProps = {
+  onPageUpdate?: (id: string, layers: string[]) => void;
+  layers?: NLSLayer[];
+  selectedPage?: Page;
+};
+export default ({ layers, selectedPage, onPageUpdate }: SettingProps) => {
+  const [layerChecked, setLayerChecked] = useState<string[]>([]);
```
KaWaite, line 12: `checkedLayers`

web/src/beta/features/Editor/Settings/hooks.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const handleAllLayersCheck = () => {
+    if (!pageId || !layers) return;
+    if (allLayersChecked) {
+      setLayerChecked([]);
+    } else {
+      const allLayerIds = layers.map(layer => layer.id);
+      setLayerChecked(allLayerIds);
+    }
+    onPageUpdate?.(pageId, allLayersChecked ? [] : layers.map(layer => layer.id));
```
KaWaite, line 57:
```suggestion
   const updatedCheckedLayers = allLayersChecked ? [] : layers.map(layer => layer.id);
   setLayerChecked(updatedCheckedLayers);
    onPageUpdate?.(pageId, updatedCheckedLayers);
```

web/src/beta/features/Editor/Settings/hooks.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  useEffect(() => {
+    if (selectedPage && layers) {
+      const allLayerIds = layers.map(layer => layer.id);
+      const allLayersSelected =
+        layerChecked.length === allLayerIds.length &&
+        layerChecked.every(layerId => allLayerIds.includes(layerId));
+      setAllLayersChecked(allLayersSelected);
+    }
+  }, [selectedPage, layers, layerChecked]);
```
KaWaite, line 30: I think you can do this simpler if you just make `allLayersChecked` a useMemo. Something like this (probably incorrect or needs some tweaking, but hopefully you get the idea):
```
const allLayersChecked = useMemo(()=> layers && checkedLayers && (layers == checkedLayers),[layers, checkedLayers]);
```

web/src/beta/features/Editor/Settings/hooks.ts
```diff
@@ -0,0 +1,67 @@
...
+    setLayerChecked(prev => {
+      const updatedLayers = prev.includes(layerId)
+        ? prev.filter(id => id !== layerId)
+        : [...prev, layerId];
+
+      onPageUpdate?.(pageId, updatedLayers);
+      if (layers) {
+        const allLayersSelected = layers.every(layer => updatedLayers.includes(layer.id));
+        setAllLayersChecked(allLayersSelected);
+      }
```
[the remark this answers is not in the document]
KaWaite: Can you check the code over for any other places where we might be doing things twice when checked layers is changed? 🙏🏻

web/src/beta/features/Editor/Visualizer/hooks.ts
mkumbobeaty: @KaWaite this filter help to obtain the layers that are available in selected page cause in a page we only have array of layersId

web/src/beta/features/Editor/Settings/index.tsx
```diff
@@ -27,22 +30,45 @@ const Settings: React.FC<Props> = ({
...
 }) => {
   const t = useT();
-  const [layerCheck, setLayerCheck] = useState(true);
+  const { allCheckedLayers, checkedLayers, handleAllLayersCheck, handleLayerCheck } = useHooks({
+    layers,
+    selectedPage,
+    onPageUpdate,
+  });
 
+  console.log(allCheckedLayers);
```
KaWaite, line 44: Remove

web/src/beta/features/Editor/Settings/hooks.ts
```diff
@@ -0,0 +1,60 @@
...
+import { NLSLayer } from "@reearth/services/api/layersApi/utils";
+import { Page } from "@reearth/services/api/storytellingApi/utils";
+
+type SettingProps = {
+  onPageUpdate?: (id: string, layers: string[]) => void;
+  layers?: NLSLayer[];
+  selectedPage?: Page;
+};
+export default ({ layers, selectedPage, onPageUpdate }: SettingProps) => {
+  const pageId = selectedPage?.id;
```
KaWaite, line 12: useMemo

web/src/beta/features/Editor/Settings/hooks.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  const allLayerIds = useMemo(() => {
+    return layers?.map(layer => layer.id) || [];
+  }, [layers]);
```
KaWaite, line 16: Since this is only used in allLayersSelected, please just put it there in that useMemo to keep code a bit more tidy.

web/src/beta/features/Editor/Settings/hooks.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  const selectedLayerIds = useMemo(() => {
+    return selectedPage?.layersIds || [];
+  }, [selectedPage]);
+
+  const allLayersSelected = useMemo(() => {
+    return selectedLayerIds.length === allLayerIds.length;
+  }, [selectedLayerIds, allLayerIds]);
```
KaWaite, line 24: This isn't guaranteed to be correct. If we have LayerA, LayerB and LayerC, then select all , then delete LayerA from the project, it'll think we haven't selected all.
I feel like maybe there is a better solution somewhere, but for now an easy fix is just to make sure `selectedLayerIds.length >= allLayerIds.length`

web/src/beta/features/Editor/Visualizer/hooks.ts
```diff
@@ -87,8 +91,14 @@ export default ({
...
-  const layers = useMemo(() => processLayers(nlsLayers), [nlsLayers]);
-
+  const layers = useMemo(
+    () =>
+      processLayers(nlsLayers)?.map(layer => ({
+        ...layer,
+        hidden: false,
+      })),
+    [nlsLayers],
+  );
```
KaWaite, line 101: On all tabs BUT story tab and publish/story we will want hidden to be how it is set in the Map tab. As far as I can tell this will force all layers to be visible from the start, which isn't good.

Then below, you have layersData state + a useEffect processing layers more, which I find these two being redundant and adding extra code.

I think better is to just do it all here in this layers useMemo.
Something like this (code is probaly not completely correct):
```suggestion
  const layers = useMemo(
    () => {
    let processedLayers = processLayers(nlsLayers)
    if (!!showStoryPanel) { 
       return processedLayers;
   };
   return processedLayers?.map(layer => ({
        ...layer,
        hidden: !currentPage?.layersIds?.includes(layer.id),
      })
     },
    [nlsLayers, currentPage?.layersIds],
  );
```

KaWaite (approved):

web/src/beta/features/Editor/Visualizer/hooks.ts
```diff
@@ -87,7 +92,16 @@ export default ({
...
 
-  const layers = useMemo(() => processLayers(nlsLayers), [nlsLayers]);
+  const layers = useMemo(() => {
+    const processedLayers = processLayers(nlsLayers);
+    if (showStoryPanel) {
+      return processedLayers;
+    }
+    return processedLayers?.map(layer => ({
+      ...layer,
+      hidden: !currentPage?.layersIds?.includes(layer.id),
```
keiya01, line 102: Why do we need `showStoryPanel`? Can't we use just `hidden` property?

mkumbobeaty: @keiya01 sure you are correct I will remove it

pyshx (approved):

web/src/beta/lib/core/Map/ClusteredLayers/index.tsx
```diff
@@ -15,6 +15,7 @@ export type Props = {
   clusters?: Cluster[];
   delegatedDataTypes?: DataType[];
   sceneProperty?: any;
+  showStoryPanel?: boolean;
```
keiya01, line 18: Can we remove this as well?

web/src/beta/lib/core/Map/Layers/index.tsx
```diff
@@ -47,6 +47,7 @@ export type Props = Omit<ClusteredLayerProps, "atomMap" | "isHidden"> & {
     info: SelectedFeatureInfo | undefined,
   ) => void;
   engineRef?: RefObject<EngineRef>;
+  showStoryPanel?: boolean;
```
keiya01, line 50: Can we remove this as well?

web/src/beta/lib/core/Map/index.tsx
```diff
@@ -27,6 +27,7 @@ export type { MapRef as Ref } from "./hooks";
 export type Props = {
   engines?: Record<string, Engine>;
   engine?: string;
+  showStoryPanel?: boolean;
```
keiya01, line 30: Can we remove this as well?

web/src/beta/lib/core/Visualizer/index.tsx
```diff
@@ -86,6 +86,7 @@ export type Props = {
   selectedWidgetArea?: WidgetAreaType;
   hiddenLayers?: string[];
   zoomedLayerId?: string;
+  showStoryPanel?: boolean;
```
keiya01, line 89: Can we remove this as well?

web/src/beta/lib/core/Map/ClusteredLayers/index.tsx
mkumbobeaty: @keiya01 this one is used to check the selected tab.For the story page we need to display layers based on the page, so if I not check which tab is selected, then when navigate to other tab it will show the layers based on  previous selected page. so showStoryPanel is need.

web/src/beta/lib/core/mantle/types/index.ts
```diff
@@ -26,12 +26,14 @@ export type LayerSimple = {
...
   defines?: Record<string, string>;
   events?: Events;
+  hidden?: boolean;
 } & Partial<LayerAppearanceTypes> &
   LayerCommon;
 
 export type LayerGroup = {
   type: "group";
   children: Layer[];
+  hidden?: boolean;
```
keiya01, line 36: Also by the way, I think we can use `visible` instead of `hidden`.
https://github.com/reearth/reearth/blob/787094e057884b96eee86982981ef9f909f17e01/web/src/beta/lib/core/mantle/types/index.ts#L43

web/src/beta/features/Editor/Visualizer/hooks.ts
KaWaite: Sorry incorrect (check slack for answer)

web/src/beta/features/Editor/Visualizer/hooks.ts
```diff
@@ -87,7 +90,13 @@ export default ({
     [selected],
   );
 
-  const layers = useMemo(() => processLayers(nlsLayers), [nlsLayers]);
+  const layers = useMemo(() => {
+    const processedLayers = processLayers(nlsLayers);
+    return processedLayers?.map(layer => ({
+      ...layer,
+      hidden: !currentPage?.layersIds?.includes(layer.id),
```
keiya01:
```suggestion
      visible: showStoryPanel && currentPage?.layersIds?.includes(layer.id),
```

keiya01 (approved):

PR #717, opened 2023-10-03
chore(web): fix upload file bug

mkumbobeaty:
# Overview

## What I've done

- fix page re-render after upload csv file in add from local

### DIFF
--- a/web/src/beta/components/Modal/index.tsx
+++ b/web/src/beta/components/Modal/index.tsx
@@ -49,11 +49,15 @@ const Modal: React.FC<Props> = ({
       }));
       setTabsFields(convertedFields);
     }
-  }, [sidebarTabs]);
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
 
   const { handleActivate, fields: tabs } = useManageSwitchState({
     fields: TabsFields,
   });
+
+  console.log("tab", tabs);
 
   const handleTabChange = useCallback(
     (tabId: string) => {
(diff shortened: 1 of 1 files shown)


web/src/beta/components/Modal/index.tsx
```diff
@@ -49,12 +49,16 @@ const Modal: React.FC<Props> = ({
       }));
       setTabsFields(convertedFields);
     }
-  }, [sidebarTabs]);
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
KaWaite, line 52: This should only be used if absolutely necessary. Can you look at how you might refactor this component to avoid this issue instead? 
For example, I am not sure setting sidebarTabs to TabsFields is necessary. Instead of local state for tabsfields, which should be static, having local state for what tab is currently selected makes more sense. Or if you can see a better implementation 🙏🏻

mkumbobeaty: @kyle sure let me check how to re-implement this

web/src/beta/components/Modal/index.tsx
```diff
@@ -39,17 +39,12 @@ const Modal: React.FC<Props> = ({
...
-  useEffect(() => {
-    if (sidebarTabs) {
-      const convertedFields: SwitchField<SidebarTab>[] = sidebarTabs.map((tab, index) => ({
-        active: index === 0,
-        ...tab,
-      }));
-      setTabsFields(convertedFields);
-    }
-  }, [sidebarTabs]);
+  const [TabsFields] = useState<SwitchField<SidebarTab>[]>(
```
KaWaite, line 42: sorry state should be lowercased. `tabsFields`

KaWaite (approved):

pyshx (approved):

PR #719, opened 2023-10-03
chore(web): refactor assets hooks

nina992:
# Overview
This pr is for refactoring the asset 
## What I've done
1. merge the two existing hooks related to asset 
2. create a hook for file uploading 
3. move hooks to where they should be used.


web/src/beta/features/Modals/ChooseAssetModal/index.tsx
```diff
@@ -185,7 +199,7 @@ const SearchButton = styled(Button)`
 const AssetListWrapper = styled.div`
   display: flex;
   flex-direction: column;
-  overflow-y: scroll;
+  overflow-y: auto;
 `;
```
airslice, line 203: Shall we set a height or max-height on this wrapper? Otherwise it will be very tall if there's a lot of assets.

web/src/beta/features/Assets/AssetHooks/hooks.tsx
KaWaite, line 1: I don't think this needs its own directory. Getting a lot of duplicate info from the path, when in reality `beta/features/Assets/hooks.ts` suffices

KaWaite: Also, hooks files typically are just `.ts` files. If possible please change the file type from tsx to ts

web/src/beta/features/Assets/AssetHooks/hooks.tsx
```diff
@@ -0,0 +1,174 @@
...
+const enumTypeMapper: Partial<Record<GQLSortType, string>> = {
+  [GQLSortType.Date]: "date",
+  [GQLSortType.Name]: "name",
+  [GQLSortType.Size]: "size",
+};
+
+function toGQLEnum(val?: SortType) {
+  if (!val) return;
+  return (Object.keys(enumTypeMapper) as GQLSortType[]).find(k => enumTypeMapper[k] === val);
+}
```
KaWaite, line 20: Is this duplicated? If so, please just use the function already there

web/src/beta/features/Assets/AssetHooks/hooks.tsx
```diff
@@ -0,0 +1,174 @@
...
+      { key: "date-reverse", label: t("First Uploaded") },
+      { key: "name", label: t("A To Z") },
+      { key: "name-reverse", label: t("Z To A") },
+      { key: "size", label: t("Size Small to Large") },
+      { key: "size-reverse", label: t("Size Large to Small") },
+    ],
+    [t],
+  );
+
+  const iconChoice =
```
KaWaite, line 74: useMemo

web/src/beta/features/Assets/AssetHooks/hooks.tsx
```diff
@@ -0,0 +1,174 @@
...
+        : "filterName"
+      : sort?.type === "size"
+      ? sort?.reverse
+        ? "filterSizeReverse"
+        : "filterSize"
+      : sort?.reverse
+      ? "filterTimeReverse"
+      : "filterTime";
+
+  const onGetMoreAssets = useCallback(async () => {
```
KaWaite, line 87: handleGetMoreAssets

web/src/beta/features/Assets/AssetHooks/hooks.tsx
```diff
@@ -0,0 +1,174 @@
...
+      await fetchMore({
+        variables: {
+          pagination: pagination(sort, endCursor),
+        },
+      });
+      isGettingMore.current = false;
+    }
+  }, [endCursor, sort, fetchMore, hasMoreAssets, isGettingMore]);
+
+  const onSortChange = useCallback(
```
KaWaite, line 99: handleSortChange

web/src/beta/features/Assets/useAssetUploader/hooks.tsx
KaWaite, line 1: Any reason these are separate from AssetHooks? Both are importing useAssetsFetcher, and what we are doing here is. Not sure its necessary to separate this only.

web/src/beta/features/Modals/ChooseAssetModal/index.tsx
KaWaite, line 1: I think features/Modals/AssetModel/index.tsx is enough. ChooseAssetModal is more info than we need, since there should be only one Asset modal

web/src/beta/features/Assets/AssetHooks/hooks.tsx
nina992: Done

web/src/beta/features/Assets/AssetHooks/hooks.tsx
nina992: sorry but where the duplicated one , it is defined only once in this file

web/src/beta/features/Assets/AssetHooks/hooks.tsx
nina992: done , note I think we won't use it , I will check with red if it will be ever used

web/src/beta/features/Assets/AssetHooks/hooks.tsx
nina992: done

web/src/beta/features/Modals/ChooseAssetModal/index.tsx
nina992: done

web/src/beta/features/Assets/AssetHooks/hooks.tsx
KaWaite: Yeah I just checked the code and it seems not to be used at the moment. If red said it'll be used, comment it out, or else delete it please 🙏🏻

nina992: I checked the design and it is not exist any more , so I removed it

web/src/beta/features/Assets/useAssetUploader/hooks.tsx
nina992: yes , as you know the upload button is not inside the asset modal in the case of url field , so if I wanted to keep it inside the general hook I have to use the hook twice  ( and  just use the upload in the parent )or use the hook  in the parent component and pass a huge amount of props to the modal , so I figure out that it could be separated and also the functionality of it can be reused easily

airslice (approved):

KaWaite (approved):

web/src/beta/features/Assets/useAssetUploader/hooks.tsx
nina992: done

PR #720, opened 2023-10-04
chore(web): update story text block fontsize options

airslice:
# Overview

This PR updates font size options of story text block and fixed a bug of lexical with solution from lexical issue page.


KaWaite (approved):

web/src/beta/lib/lexical/RichTextEditor/plugins/ToolbarPlugin.tsx
```diff
@@ -104,6 +107,51 @@ const LINE_HEIGHT_OPTIONS: [string, string][] = [
...
+  existingCssStyle: string,
+  cssProperty: string,
+  newCssValue: string,
+): string {
+  const cssArr = existingCssStyle
+    .split(";")
+    .filter(prop => !!prop)
+    .map(prop => prop.trim());
+
+  let found = false;
```
KaWaite, line 129: This doesn't seem to be needed since you return before the `if(!found){}`

PR #723, opened 2023-10-04
chore(web): refactor story dnd

mkumbobeaty:
# Overview

## What I've done

- scrolling when dragging 
- Improve drag and drop block to work when hovered


nina992 (approved):

web/src/beta/lib/core/StoryPanel/SelectableArea/index.tsx
```diff
@@ -49,15 +49,7 @@ const SelectableArea: React.FC<Props> = ({
...
-  const {
-    isHovered,
-    showPadding,
-    panelSettings,
-    setShowPadding,
-    handleMouseOver,
-    handleMouseOut,
-    handleClickAway,
-  } = useHooks({
+  const { showPadding, panelSettings, setShowPadding, handleClickAway } = useHooks({
```
KaWaite, line 52: You removed some values from the destructured hooks, but seem to not remove them from the hooks. If they aren't being used anymore, please update the hooks file too 🙏🏻

web/src/beta/components/DragAndDropList/scrollItem.tsx
```diff
@@ -1,63 +1,26 @@
...
 
-export interface IUseScroll {
-  position: number;
-  isScrollAllowed: boolean;
-}
+export const useScroll = () => {
+  const scrollContainerRef = useRef<null | HTMLDivElement>(null);
 
-const BOUND_HEIGHT = 70;
+  const handleMouseMove = (e: MouseEvent) => {
```
KaWaite, line 6: useCallback 🙏🏻

web/src/beta/components/DragAndDropList/index.tsx
```diff
@@ -25,25 +25,24 @@ function DragAndDropList<Item extends { id: string } = { id: string }>({
...
 }: Props<Item>) {
   const [movingItems, setMovingItems] = useState<Item[]>(items);
-  const listRef = useRef<null | HTMLDivElement>(null);
-
-  const { updatePosition } = useScroll(listRef);
-
+  const { scrollContainerRef } = useScroll();
   const dragDropManager = useDragDropManager();
   const monitor = dragDropManager.getMonitor();
+  const [, setIsMonitor] = useState(0);
```
KaWaite, line 31: Is this basically acting as a way to rerender the component? 
if so, can you make it clearer somehow through the name of the setter function or wrapping the setIsMonitor in a callback that is named like `handleForceRender` or something?
If I am mistaken, can you explain its purpose? 🙏🏻

mkumbobeaty: this is not causing the re-render, the monitor.subscribeToOffsetChange() is used to update the scroll position of the list when an item is being dragged, when the item is dragged, the offset changes, and the preview is updated accordingly. So if not using this subscription, the monitor might not be updated as expected because the subscription to offset changes is no longer in place after the first drag-and-drop operation. This could result in the preview(used as ref item when using custom-dnd-handler) not being updated correctly after the first drag-and-drop operation.

KaWaite (approved):

web/src/beta/lib/core/StoryPanel/SelectableArea/hooks.ts
```diff
@@ -43,12 +27,9 @@ export default ({ editMode, isSelected, property, setEditMode, onClickAway }: Pr
   }, [property?.panel]);
 
   return {
-    isHovered,
     showPadding,
     panelSettings,
```
KaWaite, line 31: Can you delete using sending panelSettings here since its not being used anymore 🙏🏻 After that you can close this comment 👍🏻

PR #727, opened 2023-10-06
chore(web): show installed widgets & fix computed feature property

airslice:
# Overview

This PR:
- Allows show installed widgets.
- Fixs the computed feature property.
- Comment the setting id for legacy layer to avoid crash.

**THIS IS REQUIRED FOR SHOW CASE.**

### DIFF
--- a/web/src/beta/lib/core/Map/Layers/hooks.ts
+++ b/web/src/beta/lib/core/Map/Layers/hooks.ts
@@ -365,11 +365,11 @@ export default function useHooks({
         rawLayer.type === "simple" &&
         rawLayer.data?.value &&
         // If data isn't cachable, reuse layer id for performance.
-        DATA_CACHE_KEYS.some(k => !rawLayer.data?.[k])
+        DATA_CACHE_KEYS.some(k => !rawLayer.data?.[k]) &&
+        Object.isExtensible(rawLayer.data.value)
       ) {
         // If layer property is overridden, feature is legacy layer.
         // So we can set layer id to prevent unnecessary render.
-        // TODO: fix: Cannot add property id, object is not extensible
         rawLayer.data.value.id = id;
       }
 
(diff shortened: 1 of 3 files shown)


web/src/services/api/widgetsApi/utils.ts
```diff
@@ -40,8 +40,9 @@ export const getInstallableWidgets = (rawScene?: GetSceneQuery) => {
       return plugin?.extensions
         .filter(
           e =>
-            e.type === PluginExtensionType.Widget &&
-            AVAILABLE_WIDGET_IDS.includes(`reearth/${e.extensionId}`),
+            (e.type === PluginExtensionType.Widget &&
+              AVAILABLE_WIDGET_IDS.includes(`reearth/${e.extensionId}`)) ||
+            plugin.id !== "reearth",
```
KaWaite, line 45:
```
(e.type === PluginExtensionType.Widget &&
              (AVAILABLE_WIDGET_IDS.includes(`reearth/${e.extensionId}`)) ||
            plugin.id !== "reearth")
```

KaWaite: This is the widgetsApi, so should be widget extension 👍🏻

airslice: Updated!

web/src/beta/lib/core/Map/Layers/hooks.ts
```diff
@@ -365,7 +365,8 @@ export default function useHooks({
         rawLayer.type === "simple" &&
         rawLayer.data?.value &&
         // If data isn't cachable, reuse layer id for performance.
-        DATA_CACHE_KEYS.some(k => !rawLayer.data?.[k])
+        DATA_CACHE_KEYS.some(k => !rawLayer.data?.[k]) &&
+        Object.isExtensible(rawLayer.data.value)
```
keiya01, line 369: User need to take care not to pass inextensible object if some problem occurs later.

KaWaite (approved):

PR #740, opened 2023-10-16
chore(web): add date time field component

nina992:
# Overview
Add Date and Time for beta reearth 
## What I've done
image
image


mkumbobeaty (approved):

web/src/beta/components/Icon/Icons/calender.svg
KaWaite, line 1: currentColor

web/src/beta/components/Icon/Icons/clock.svg
KaWaite, line 1: currentColor

web/src/beta/components/fields/common/TextInput/index.tsx
```diff
@@ -1,12 +1,14 @@
 import { useCallback, useEffect, useRef, useState } from "react";
 
 import { styled } from "@reearth/services/theme";
+//import "./index.module.css";
```
KaWaite, line 4: remove

web/src/beta/components/fields/common/TextInput/index.tsx
```diff
@@ -87,4 +91,5 @@ const StyledInput = styled.input`
   :focus {
     border-color: ${({ theme }) => theme.outline.main};
   }
+  color-scheme: dark;
```
KaWaite, line 94: We have never used color-scheme before, so is there a reason you are using it here instead of relying on the background and color using our theme? 🤔

web/src/beta/components/fields/common/TextInput/index.tsx
```diff
@@ -1,12 +1,14 @@
...
 
 import { styled } from "@reearth/services/theme";
+//import "./index.module.css";
 
 export type Props = {
   value?: string;
   placeholder?: string;
   timeout?: number;
   autoFocus?: boolean;
+  type?: string;
```
KaWaite, line 11: type?: HTMLInputTypeAttribute

web/src/beta/components/fields/DateTimeField/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+    if (value) {
+      const [dateString, timeString] = value.split(" ");
+      setTime(timeString);
+      setDate(dateString);
+    }
+  }, [value]);
+
+  useEffect(() => {
+    if (value) onChange?.(date + " " + time);
+  }, [date, onChange, time, value]);
```
KaWaite, line 29: This is going to create an infinite loop saving the same value over and over again. 
I think what you were wanting to do was
```
useEffect(() => {
onChange?.(date + " " + time);
  }, [date, onChange, time]);
```
which would call onChange for any change of date or time. But I do think we are handling onChanges in other field components through the onChange/Click calllback.
I feel like doing mutations through the useEffect should be only very rarely used since we have less control over a useEffect and the chances of sideeffects like infinite loops increase significantly

web/src/beta/components/fields/DateTimeField/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  return (
+    <Property name={name} description={description}>
+      <Wrapper>
```
KaWaite, line 33: The `Property` component already is flexing, so I don't think you need this Wrapper component 🙏🏻

web/src/beta/components/Icon/Icons/calender.svg
nina992: done

web/src/beta/components/Icon/Icons/clock.svg
nina992: done

web/src/beta/components/fields/common/TextInput/index.tsx
nina992: done

web/src/beta/components/fields/common/TextInput/index.tsx
nina992: yes , as you noticed I used the default html calendar and clock and in order to re-style it I add this attribute so it will have these colors  , and this is the only way to do that or we need to work on all -webkit classes( which I tried to find all of them and I needed to add a whole css file inorder to change them and even though the style was broken) , but this way only set the scheme to be dark and then all what I need is done.

web/src/beta/components/fields/common/TextInput/index.tsx
nina992: done

web/src/beta/components/fields/DateTimeField/index.tsx
nina992: No it flex as rows , I need one row here

web/src/beta/components/fields/DateTimeField/index.tsx
nina992: Actually I thought of it like this  because I have to have the two values together (date , time)

nina992: done

web/src/beta/components/Icon/Icons/calender.svg
KaWaite: currentcolor for stroke values, not fill 🙏🏻

web/src/beta/components/Icon/Icons/clock.svg
KaWaite: currentcolor for stroke values, not fill 🙏🏻

web/src/beta/components/fields/common/TextInput/index.tsx
KaWaite: Ok, I guess as a temp solution this is fine 🙏🏻

KaWaite (approved):

PR #741, opened 2023-10-17
chore(web): Updated fields in Settings

jashanbhullar:
# Overview
- Property Fields in settings Page

## What I've done
- Added textaread field 
- textinput now supports disabled flag 
- Updated fields in settings page 

## What I haven't done
- NA

## How I tested
- UI


web/src/beta/features/ProjectSettings/innerPages/StorySettings/StorySettingsDetail.tsx
```diff
@@ -16,18 +18,33 @@ type Props = {
...
   }, [localPanelPosition, onUpdateStory]);
 
+  const options = useMemo(
+    () =>
+      Object.keys(Position).map(label => ({
+        label,
+        key: Position[label as keyof typeof Position],
+      })),
+    [],
+  );
```
KaWaite, line 37: I don't think this will support internationalization. Can you update this code to include it please 🙏🏻

KaWaite (approved):

web/src/beta/features/ProjectSettings/innerPages/StorySettings/StorySettingsDetail.tsx
jashanbhullar: So when i tried 
```
label: t(label) 
```
i18n didn't pick up the translation probably because it wasn't string so I just used a variable rather than this loop.

PR #745, opened 2023-10-18
chore(web): refactor style for some beta components

nina992:
# Overview
This Pr is for refactoring some styles to match the design  
## What I've done
- Refactor the Data Source manger fields : url field and select field have wrong background color 
image
- Refactor Asset modal: fix scroll bar, fix cards dimension and gaps 
image

### DIFF
--- a/web/src/beta/features/Assets/AssetCard/index.tsx
+++ b/web/src/beta/features/Assets/AssetCard/index.tsx
@@ -46,8 +46,9 @@ const Wrapper = styled.div<{ selected?: boolean }>`
   border: 2px solid ${({ selected, theme }) => (selected ? `${theme.select.main}` : "transparent")};
   padding: ${({ theme }) => theme.spacing.small}px;
   width: 100%;
-  max-width: 148px;
+  max-width: 120px;
   height: 100%;
+  max-height: 103px;
   position: relative;
   cursor: pointer;
   color: ${({ theme }) => theme.content.main};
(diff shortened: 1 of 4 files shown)


web/src/beta/features/Assets/AssetCard/index.tsx
```diff
@@ -46,8 +46,9 @@ const Wrapper = styled.div<{ selected?: boolean }>`
   border: 2px solid ${({ selected, theme }) => (selected ? `${theme.select.main}` : "transparent")};
   padding: ${({ theme }) => theme.spacing.small}px;
   width: 100%;
-  max-width: 148px;
+  max-width: 120px;
   height: 100%;
+  max-height: 103px;
```
KaWaite, line 51: Should we be doing max-width/height? Or just setting the width and height to those values?

web/src/beta/features/Modals/AssetModal/index.tsx
```diff
@@ -242,15 +241,32 @@ const AssetListWrapper = styled.div`
...
+  }
+  ::-webkit-scrollbar-track {
+    background-color: transparent;
+    background-color: red;
+    display: none;
+  }
+  ::-webkit-scrollbar-thumb {
+    border-radius: 5px;
+    background-color: ${({ theme }) => theme.bg[2]};
+  }
```
KaWaite, line 261: I think the scrollbar is taking up space so when it is not showing it looks like there is more padding on the right than the left

web/src/beta/features/Assets/AssetCard/index.tsx
nina992: I think it is better to set max values , since we did that with div container

PR #755, opened 2023-10-23
chore(web):  add scene setting

nina992:
# Overview
Add Scene Setting to the Editor for beta Project 
## What I've done
image
image
image
image
image
image
image

## Memo
Design Link : https://www.figma.com/file/W6nDSiCBLubcNeAT6RkOoa/%E2%AD%90%E2%AD%90%E2%AD%90Re%3AEarth-Beta-Design-system?type=design&node-id=2811-10042&mode=design&t=D9eWg6IuRUjYq5nh-0


web/src/beta/features/Editor/tabs/map/RightPanel/index.tsx
```diff
@@ -46,7 +48,10 @@ const MapRightPanel: React.FC<Props> = ({
   const { scene } = useSceneQuery({ sceneId });
 
   const scenePropertyId = useMemo(() => scene?.property?.id, [scene?.property?.id]);
-  const sceneSettings = useMemo(() => convert(scene?.property), [scene?.property]);
+  const sceneSettings = useMemo(
+    () => convert(scene?.property)?.filter(item => item.collection === selectedSceneSetting),
```
KaWaite, line 52: Can you add the converting of the scene property to inside the scene API and update the scene type there?
Currently we are asserting the type as the GQL type, which isn't great, so if you can update the type too that would be really helpful 🙏🏻

KaWaite: Of course the filtering here should stay here

web/src/beta/features/Editor/index.tsx
```diff
@@ -132,8 +133,9 @@ const Editor: React.FC<Props> = ({ sceneId, projectId, workspaceId, tab }) => {
     nlsLayers,
     currentPage,
     currentCamera,
-    showSceneSettings: selectedSceneSetting,
+    showSceneSettings: !!selectedSceneSetting,
```
KaWaite, line 136: This isn't needed anymore. This can be achieved by just checking if selectedSceneSetting exists inside of the left panel

web/src/beta/features/Editor/index.tsx
```diff
@@ -75,6 +74,7 @@ const Editor: React.FC<Props> = ({ sceneId, projectId, workspaceId, tab }) => {
     isVisualizerReady,
     visualizerRef,
   });
+  const { scene, selectedSceneSetting, handleSceneSettingSelect } = useScene({ sceneId });
```
KaWaite, line 77: You've set up this useScene, but then only pass the scene to the left panel, and then still use the right panel's direct use of the scene api to fetch the scene. 
If you want to consolidate scene in this useScene, please at least update the files you are working in in this PR to match.

Aka, if useLeftPanel will take a scene prop and not access the scene api directly, make sure the useRightPanel is updated accordingly since you are already modifying both in this PR.

web/src/beta/features/Editor/index.tsx
nina992: done

web/src/beta/features/Editor/index.tsx
nina992: Done

web/src/beta/features/Editor/tabs/map/RightPanel/index.tsx
nina992: sorry would you explain more , you mean move this convert to useScene, not scene Api right? and this scene setting have the type Item not GQL type . is that the type you are talking about?

KaWaite: If you check other useXxxxApi's and how they are doing properties, you will see they should all be doing `convert(asldkfj?.property)` there.
Just want to keep it consistent. But this also means you'll need to update/add a Scene type there

nina992: I apply your suggestion ,but I didn't need to define type , but can take a look is that what you mean ?

PR #759, opened 2023-10-24
chore(server, web): add map valueType to property

pyshx:
# Overview
This PR adds support for `map` valueType for property to the backend, value `Map` has been utilized from `gqlgen` to provide support for this type. The appropriate changes has also been made in the frontend by `yarn gql`.


KaWaite (approved):

server/pkg/builtin/manifest.yml
```diff
@@ -2544,7 +2544,7 @@ extensions:
               title: Background Color
               ui: color
             - id: showLayers
-              type: string
+              type: map
```
jashanbhullar, line 2547: image
The design has been updated, can you add another property with same type as map - hideLayers?

pyshx: we'll add it in the next PR.

PR #762, opened 2023-10-24
chore: fix page title empty string

nina992:
# Overview
Page title can become empty string and then can't be selected anymore (should go back to UNTITLED)


web/src/beta/lib/core/StoryPanel/Block/builtin/Title/index.tsx
```diff
@@ -24,6 +24,7 @@ const TitleBlock: React.FC<Props> = ({ block, isSelected, ...props }) => {
     () => property?.title?.color?.value as ValueTypes["string"],
     [property?.title?.color?.value],
   );
+  const hasEmptySpace = /^ *$/.test(title);
```
KaWaite, line 27: I don't think you need to do this. All you should need to do is change `{title ?? t("Untitled")}` to `{title || t("Untitled")}` 👍🏻

web/src/beta/components/fields/common/TextInput/index.tsx
```diff
@@ -32,6 +32,14 @@ const TextInput: React.FC<Props> = ({
     setCurrentValue(value ?? "");
   }, [value]);
 
+  const handlepdateValue = useCallback(
+    (newValue: string) => {
+      if (!newValue || /^ *$/.test(newValue)) setCurrentValue("");
+      onChange?.(newValue);
+    },
+    [onChange],
+  );
```
KaWaite, line 41: Not quite sure what this is doing. Why do we want to in some instances set currentValue to an empty string but then pass newValue to onChange?

web/src/beta/lib/core/StoryPanel/Block/builtin/Title/index.tsx
nina992: done

web/src/beta/components/fields/common/TextInput/index.tsx
nina992: the idea here that some times the user enter spaces , so I don't want to update with this spaces instead only save empty space

web/src/beta/lib/core/StoryPanel/Block/builtin/Title/index.tsx
KaWaite: I just tested myself and it works 🤔

web/src/beta/components/fields/common/TextInput/index.tsx
KaWaite: This isn't really good because it alters what the user can expect without any notice.
Also, this is on the common TextInput component, which in some cases a user might want to add white space. One example is currently we don't allow align center to the page title, so they might want to center the title with spacing. I feel like that should be allowed.

nina992: well since it is acceptable for system , so I will just remove those changes here

web/src/beta/lib/core/StoryPanel/Block/builtin/Title/index.tsx
nina992: https://github.com/reearth/reearth/assets/89770889/5fe37970-490e-4b0a-a9da-27f2bb10afbc

KaWaite: Ah sorry I misunderstood what you meant...

web/src/beta/features/Editor/tabs/story/LeftPanel/ContentPage/index.tsx
```diff
@@ -60,6 +60,7 @@ const ContentPage: React.FC<Props> = ({
           renderItem={(storyPage, i) => {
             const title = (getFieldValue(storyPage.property.items ?? [], "title", "title") ??
               t("Untitled")) as string;
+            const hasEmptySpace = /^ *$/.test(title);
```
KaWaite, line 63: Can you move this in to a global util so we can use it elsewhere. Something like [this](https://stackoverflow.com/questions/57476487/how-to-check-if-a-string-contains-only-white-space-or-is-empty-in-typescript#:~:text=export%20function%20isEmpty(text%3A%20string)%3A%20boolean%20%7B%0A%20%20return%20text%20%3D%3D%20null%20%7C%7C%20text.match(/%5E%5Cs*%24/)%20!%3D%3D%20null%3B%0A%7D) might be more useable? Maybe

KaWaite (approved):

web/src/beta/features/Editor/tabs/story/LeftPanel/ContentPage/index.tsx
nina992: done

PR #767, opened 2023-10-25
chore(web): date time field input

mkumbobeaty:
# Overview

## What I've done
- Improved date time field input


web/package.json
```diff
@@ -152,6 +152,7 @@
         "lodash-es": "4.17.21",
         "lru-cache": "8.0.4",
         "mini-svg-data-uri": "1.4.4",
+        "moment-timezone": "^0.5.43",
```
KaWaite, line 155: Please remove the ^ and do yarn again 🙏🏻 If we don't, everyone's local can have different versions and it gets hard to troubleshoot local issues

web/src/services/i18n/translations/ja.yml
```diff
@@ -18,6 +18,11 @@ Edit: 編集
 Color Picker: カラーピッカー
 You have passed the maximum value.: You have passed the maximum value.
 You have passed the minimum value.: You have passed the minimum value.
+Set Time: ''
```
KaWaite, line 21: 時間設定

web/src/services/i18n/translations/ja.yml
```diff
@@ -18,6 +18,11 @@ Edit: 編集
...
+Date: 作成日時順
+Time: ''
```
KaWaite, line 23: 時間

web/src/services/i18n/translations/ja.yml
```diff
@@ -18,6 +18,11 @@ Edit: 編集
...
+Time Zone: ''
```
KaWaite, line 24: 時間帯

web/src/services/i18n/translations/en.yml
```diff
@@ -18,6 +18,11 @@ Edit: Edit
 Color Picker: Color Picker
 You have passed the maximum value.: You have passed the maximum value.
 You have passed the minimum value.: You have passed the minimum value.
+Set Time: ''
+Date: Date
+Time: ''
+Time Zone: ''
+set: ''
```
KaWaite, line 25:
```suggestion
Set Time: Set Time
Date: Date
Time: Time
Time Zone: Time Zone
set: set
```

web/src/beta/components/fields/DateTimeField/EditPanel/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+        <TextWrapper>
+          <Label>{t("Time")}</Label>
+
+          <TextInput className="customTextInput" type="time" value={time} onChange={onTimeChange} />
+        </TextWrapper>
+        <SelectWrapper>
+          <Label>{t("Time Zone")}</Label>
+          <SelectField
+            value={selectedTimezone}
+            className="timezone"
```
KaWaite, line 50: Is this timezone className doing anything? Since we aren't using any external css files (and you aren't adding any in this PR) that would define a timezone class seems like its maybe a mistake?

web/src/beta/components/fields/DateTimeField/EditPanel/hooks.ts
```diff
@@ -0,0 +1,77 @@
...
+      const offsetA = parseInt(a.offset);
+      const offsetB = parseInt(b.offset);
+      return offsetA - offsetB;
+    });
+
+    return uniqueTimezones;
+  }, []);
+
+  const zones: string[] = moment.tz.names();
+  const offsetFromUTC: TimezoneInfo[] = getUniqueTimezones(zones);
```
KaWaite, line 56: If I am wrong please let me know, but I feel like this will be relatively static, so can you pull these out of the hook? If moment must be used inside a hook or a react component, maybe its fine here. But it seems more like a util that we can import and use once in a useMemo and then it shouldn't re-render at all and keeps the code simpler.

Whether or not you do the above, also please:
- put `const zones: string[] = moment.tz.names();` inside getUniqueTimezones because its not being used anywhere else
- make offsetFromUTC a useMemo

web/src/beta/components/fields/DateTimeField/EditPanel/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+    selectedTimezone,
+    onDateChange,
+    onTimeChange,
+    getUniqueTimezones,
+    onTimezoneSelect,
+    onDateTimeApply,
+  } = useHooks({ onChange });
+
+  const isButtonDisabled = date.trim() === "" || time.trim() === "";
+  const uniqueTimezones = getUniqueTimezones(timezones);
```
KaWaite, line 32: You are already doing this in the hooks with `offsetFromUTC`. Just pass that from the hooks and use that instead.

web/src/beta/components/fields/DateTimeField/EditPanel/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+    timezones,
+    selectedTimezone,
+    onDateChange,
+    onTimeChange,
+    getUniqueTimezones,
+    onTimezoneSelect,
+    onDateTimeApply,
+  } = useHooks({ onChange });
+
+  const isButtonDisabled = date.trim() === "" || time.trim() === "";
```
KaWaite, line 31: Move in to hooks and wrap in useMemo please

web/src/beta/components/fields/DateTimeField/EditPanel/index.tsx
```diff
@@ -0,0 +1,116 @@
+import Button from "@reearth/beta/components/Button";
+import PanelCommon from "@reearth/beta/components/fields/CameraField/PanelCommon";
```
KaWaite, line 2: Please move this in to fields/common if its not just a CameraField related thing anymore 🙏🏻

web/src/beta/components/fields/SelectField/index.tsx
```diff
@@ -22,6 +22,7 @@ export type Props = {
   // Property field
   name?: string;
   description?: string;
+  className?: string;
```
KaWaite, line 25: Since you aren't using the `timezone` className is this change still necessary? Can you clean up any parts related to that that aren't needed? 🙏🏻

mkumbobeaty: @KaWaite  I come to realize that className is used to override the `PickerWrapper` height and add scroll, I removed this but the select in timezone look ugly, if you may have any suggestion to add the height, overflow and with in this component.

nina992 (approved):

KaWaite (approved):

PR #775, opened 2023-11-01
chore(web): Add story background setting

nina992:
# Overview
In This  pr we give the user the ability to modify the color of story background. 
## What I've done
- Add color field to story setting 
- Modify the gql query 
- Make sure the story bgColor modified in/out editor mode 
image
image


web/src/beta/features/ProjectSettings/innerPages/StorySettings/StorySettingsDetail.tsx
```diff
@@ -63,3 +73,16 @@ const StorySettingsDetail: React.FC<Props> = ({ settingsItem, onUpdateStory }) =
...
+  justify-content: space-between;
+  flex-direction: row;
+  gap: 8px;
+`;
+const SelectFieldWrapper = styled(SelectField)`
+  width: 100%;
+`;
+const ColorFieldWrapper = styled(ColorField)`
+  width: 100%;
+`;
```
KaWaite, line 88: I feel like this shouldn't be necessary. The field's Property component that all fields share should already be taking up 100% width of the the parent (I think through flexing)...

nina992: yes I thought that too, but the style was broken so I  make this change here because those components are used in alot of UIs and to avoid destroy any design.

web/src/beta/lib/core/StoryPanel/index.tsx
```diff
@@ -114,8 +113,8 @@ export const StoryPanel = memo(
 
 export default StoryPanel;
 
-const PanelWrapper = styled.div`
+const PanelWrapper = styled.div<{ bgColor?: string }>`
   flex: 0 0 ${STORY_PANEL_WIDTH}px;
-  background: #f1f1f1;
+  background: ${({ bgColor }) => (bgColor ? bgColor : "#f1f1f1")};
```
KaWaite, line 118: Can keep things a little simpler/cleaner like this
```suggestion
  background: ${({ bgColor }) => bgColor || "#f1f1f1"};
```

web/src/beta/features/ProjectSettings/innerPages/StorySettings/index.tsx
```diff
@@ -54,7 +54,10 @@ const StorySettings: React.FC<Props> = ({
         ) : (
           <StorySettingsDetail
             key={currentStory.id}
-            settingsItem={currentStory}
+            settingsItem={{
+              ...currentStory,
+              bgColor: currentStory.bgColor as string,
```
KaWaite, line 59: I feel like if this is necessary typings are wrong somewhere 🤔

web/src/beta/features/Editor/Visualizer/convert-story.ts
```diff
@@ -41,6 +41,7 @@ export const convertStory = (story?: GqlStory): Story | undefined => {
     id: story.id,
     title: story.title,
     position: story.panelPosition === "RIGHT" ? "right" : "left",
+    bgColor: story.bgColor ?? "#f1f1f1",
```
KaWaite, line 44: You also have the fallback `#f1f1f1` being done in the CSS. Just set bgColor here or in the CSS, not both. (preferably take a look at padding/gap/other fallbacks and see where we set them and do bgColor there too 🙏🏻

web/src/beta/lib/core/StoryPanel/index.tsx
nina992: I will remove it all because it is set on convert story

web/src/beta/features/Editor/Visualizer/convert-story.ts
nina992: done, I will keep it here

web/src/beta/features/ProjectSettings/innerPages/StorySettings/index.tsx
nina992: I added this typing because I faced this error "Type 'Maybe<string> | undefined' is not assignable to type 'string | undefined'.
  Type 'null' is not assignable to type 'string | undefined'" so also to make sure the bgColor has default value , do you have any suggestion here ?

KaWaite: If you are getting `Type 'null' is not ...` errors that means you need to either add it to the type, or do something like `someValue || undefined` so that if someValue is NULL it'll become undefined, not NULL

web/src/beta/components/fields/SelectField/index.tsx
```diff
@@ -51,7 +53,7 @@ const SelectField: React.FC<Props> = ({
   }, [options, value]);
 
   return (
-    <Property name={name} description={description}>
+    <Property name={name} description={description} className={className}>
```
mkumbobeaty, line 56: @KaWaite is it okay/necessary to have class here? I thought we are avoiding to use them?

KaWaite (approved):

web/src/beta/components/fields/SelectField/index.tsx
nina992: I used it because I wanted to modify the width as styled component and to avoid modify the style here immediately as we are using it widely in system

mkumbobeaty (approved):

web/src/beta/components/fields/SelectField/index.tsx
mkumbobeaty: @nina okay approved

PR #782, opened 2023-11-01
chore(web): visualizer clock update 

mkumbobeaty:
# Overview

## What I've done

- update visualizer clock when time is set in a page


web/src/beta/lib/core/StoryPanel/hooks.ts
```diff
@@ -61,6 +63,19 @@ export default (
...
+      const timePointField = page.property?.timePoint;
+      const currentTime = timePointField?.timePoint?.value;
+      if (!currentTime) onTimeChange?.(new Date());
+      else {
+        const getNewDate = new Date(currentTime.substring(0, 19)).getTime();
+        return onTimeChange?.(new Date(getNewDate));
+      }
+    },
+    [onTimeChange],
+  );
```
KaWaite, line 77: If I understand the logic here correctly, are we changing the time to the actual time/date on page change if there is no `timePoint?.value`? Hmmm maybe something to discuss with @lavalse

web/src/beta/lib/core/StoryPanel/hooks.ts
```diff
@@ -61,6 +63,19 @@ export default (
...
   );
 
+  const handlePageTime = useCallback(
+    (page: StoryPage) => {
+      const timePointField = page.property?.timePoint;
+      const currentTime = timePointField?.timePoint?.value;
+      if (!currentTime) onTimeChange?.(new Date());
+      else {
+        const getNewDate = new Date(currentTime.substring(0, 19)).getTime();
+        return onTimeChange?.(new Date(getNewDate));
```
KaWaite, line 73: Is this correct? Why do you have to pass `getNewDate` in to another `new Date()` when you created getNewDate with a `new Date()`? If you don't do `getTime()` won't you have what you want?

web/src/beta/features/Editor/useStorytelling.ts
```diff
@@ -42,6 +43,21 @@ export default function ({ sceneId }: Props) {
     [selectedStory?.pages, selectedStoryPageId],
   );
 
+  const onTimeChange = useCallback(
```
KaWaite, line 46: If you use this onTimeChange in the right panel like you are doing, this is fine.
But you also pass this in to the Visualizer, which is strange. 
Use of the visualizerRef should be to use features inside of core/Visualizer that you normally wouldn't be able to use outside. But setting upthis function and then pass it to the Visuzlier doesn't make much sense.

web/src/beta/components/fields/PropertyFields/index.tsx
KaWaite, line 1: I think this file got recreated when you were resolving merge conflicts. This file was deleted. If you made any changes to this file please move those changes to either PropertyField or PropertyItem or PropertyList, whichever it applies to

web/src/beta/components/fields/DateTimeField/index.tsx
```diff
@@ -27,7 +28,8 @@ const DateTimeField: React.FC<Props> = ({ name, description, value, onChange })
     if (!value) return;
     setDateTime("");
     onChange?.();
-  }, [value, onChange]);
+    onTimeChange?.(new Date());
```
KaWaite, line 31: Do we actually want to change the Visualizer's time here?

web/src/beta/components/fields/DateTimeField/EditPanel/hooks.ts
```diff
@@ -44,8 +45,10 @@ export default ({ value, onChange, setDateTime }: Props) => {
       const formattedDateTime = `${date}T${time}:00${selectedTimezoneInfo.offset}`;
       setDateTime?.(formattedDateTime);
       onChange?.(formattedDateTime);
+      const t = new Date(formattedDateTime.substring(0, 19)).getTime();
+      onTimeChange?.(new Date(t));
```
KaWaite, line 49: Again, are we sure we want to change the viz's time here?

web/src/beta/lib/core/StoryPanel/hooks.ts
KaWaite, line 77: I think the logic you have in the onTimeChange inside of the Editor should be here.
Like the flyTo, I think the StoryPanel should handle the logic without core's parent having to worry about it.

web/src/beta/components/fields/PropertyFields/index.tsx
mkumbobeaty: @KaWaite sure

web/src/beta/features/Editor/useStorytelling.ts
```diff
@@ -42,6 +43,21 @@ export default function ({ sceneId }: Props) {
...
+        payload: {
+          start: visualizerRef?.current?.timeline?.current?.computedTimeline?.start,
+          current: time,
+          stop: visualizerRef?.current?.timeline?.current?.computedTimeline?.stop,
+        },
+        committer: { source: "storyPage", id: currentPage?.id },
+      });
+    },
+    [currentPage?.id, visualizerRef],
+  );
```
KaWaite, line 59: This doesn't have anything to do with story anymore (it'll be used in field components only, right?). So please move it somewhere appropriate and make sure the commit is appropriate too (right now you are setting the committer to storyPage, which again isn't accurate)

KaWaite (approved):

PR #790, opened 2023-11-06
chore(web): add height to marker appearance

pyshx:
# Overview
This PR adds support for `height` as a possible parameter for marker's appearance. After this change the user can directly add and adjust `height` through the style code.

### DIFF
--- a/web/src/beta/lib/core/engines/Cesium/Feature/Marker/index.tsx
+++ b/web/src/beta/lib/core/engines/Cesium/Feature/Marker/index.tsx
@@ -29,7 +29,7 @@ export default function Marker({ property, id, isVisible, geometry, layer, featu
     () =>
       geometry?.type === "Point"
         ? property?.height
-          ? [...geometry.coordinates, property.height]
+          ? [...geometry.coordinates.slice(0, 2), property.height]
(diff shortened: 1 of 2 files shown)


web/src/beta/lib/core/engines/Cesium/Feature/Marker/index.tsx
```diff
@@ -28,7 +28,9 @@ export default function Marker({ property, id, isVisible, geometry, layer, featu
   const coordinates = useMemo(
     () =>
       geometry?.type === "Point"
-        ? geometry.coordinates
+        ? property?.height
+          ? [...geometry.coordinates, property.height]
```
keiya01, line 32: We should do like this, because`geometry.coordinates` might have height in the index of 3.
```suggestion
          ? (property.height ? [...geometry.coordinates.slice(0, 2), property.height] : geometry.coordinates)
```

keiya01 (approved):

web/src/beta/lib/core/engines/Cesium/Feature/Marker/index.tsx
pyshx: Done.

PR #806, opened 2023-11-13
chore(web): Story page click away

mkumbobeaty:
# Overview

## What I've done

- Remove story page clickaway


web/src/beta/lib/core/StoryPanel/SelectableArea/index.tsx
```diff
@@ -55,49 +53,44 @@ const SelectableArea: React.FC<Props> = ({
...
-          onRemove={onRemove}
-          onPropertyUpdate={onPropertyUpdate}
-          onPropertyItemAdd={onPropertyItemAdd}
-          onPropertyItemMove={onPropertyItemMove}
-          onPropertyItemDelete={onPropertyItemDelete}
-        />
-        {children}
-      </Wrapper>
-    </ClickAwayListener>
+    <Wrapper isSelected={isSelected} noBorder={noBorder}>
```
KaWaite, line 71: I don't think you should have removed the onClick on the Wrapper. Now the only way to select the Page is on the empty space below the page, not the page itself.

KaWaite (approved):

PR #808, opened 2023-11-13
chore(web): modify zoom level field

nina992:
# Overview


web/src/beta/components/fields/Property/PropertyField/index.tsx
```diff
@@ -154,6 +155,18 @@ const PropertyField: React.FC<Props> = ({
           onSave={handleChange}
           onFlyTo={onFlyTo}
         />
+      ) : schema.type === "array" ? (
+        schema.ui === "range" && (
```
airslice, line 159: For me i prefer `schema.type ===  "array" && schema.ui === "range" ? <RangeField/> : <p/>` here. In this way it's easier for developer to notice if he add some schema in type of "array" but no ui / has other ui.

lavalse:
image
If possible, for UI side, the default range should be full ( 0~30 )

web/src/beta/components/RangeSlider/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+const RangeSliderWithTooltip = RCSlider.createSliderWithTooltip(RCSlider.Range);
+
+export type Props = {
+  min?: number;
+  max?: number;
+} & ComponentProps<typeof RangeSliderWithTooltip>;
+
+const RangeSlider: React.FC<Props> = ({ ...props }) => {
+  const calculatedStep = props.step ? props.step : props.max ? props.max / 10 : 0.1;
```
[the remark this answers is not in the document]
nina992: Done

web/src/beta/components/fields/Property/PropertyField/index.tsx
KaWaite: You said done, but it doesn't seem like you have made these changes yet?

web/src/beta/components/RangeSlider/index.tsx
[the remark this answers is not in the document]
KaWaite: Same here

web/src/beta/lib/core/engines/Cesium/core/Imagery.tsx
```diff
@@ -45,13 +44,13 @@ export default function ImageryLayers({ tiles, cesiumIonAccessToken }: Props) {
...
-        .map(({ id, tile_opacity: opacity, tile_minLevel: min, tile_maxLevel: max, provider }, i) =>
+        .map(({ id, tile_opacity: opacity, tile_zoomLevel, provider }, i) =>
           provider ? (
             <ImageryLayer
               key={`${id}_${i}_${counter.current}`}
               imageryProvider={provider}
-              minimumTerrainLevel={min}
-              maximumTerrainLevel={max}
+              minimumTerrainLevel={tile_zoomLevel?.[0] ?? 0}
+              maximumTerrainLevel={tile_zoomLevel?.[1] ?? 30}
```
KaWaite, line 53: We aren't handling any other value's fallbacks at this level. can you set the. `?? 0` `?? 30` where we are processing tile_zoomLevel? 🙏🏻

web/src/beta/components/fields/RangeField/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+import Property from "@reearth/beta/components/fields";
+import RangeSlider, { Props as RangeProps } from "@reearth/beta/components/RangeSlider";
+
+export type Props = {
+  name?: string;
+  description?: string;
+} & RangeProps;
+
+const RangeField: React.FC<Props> = ({ name, description, value, onChange, ...args }: Props) => {
+  const [internalState, setInternalState] = useState<number[] | undefined>(value ?? [0, 30]);
```
KaWaite, line 12: RangeField isn't always going to be used for tile's zoom levels, so the [0, 30] isn't appropriate here.
If `args` includes min/max, you could destructure those from the props and use those as a starting value.

web/src/beta/lib/core/engines/Cesium/core/Imagery.tsx
nina992: I think there is no need for process it  as it will be set by the defaultValue, I will just remove it

web/src/beta/components/fields/RangeField/index.tsx
nina992: I also remove this value

web/src/beta/components/RangeSlider/index.tsx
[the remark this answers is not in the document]
airslice: Hi @nina992 i notice you updated this part with `!!min && !!max`. Maybe it's better to check value like `min !== undefined` since `!!0` will be `false` and `0` could be widely used for min.

web/src/beta/components/fields/Property/PropertyField/index.tsx
airslice: Emmm i'm not sure whether i'm reading the latest code or not... but basically what i mentioned is I guess it's better to display a `<p key={schema.id}>{schema.name} field</p>` instead of nothing when a data with `type` "array" but no `ui` (or `ui` is not range) comes. Currently it will still become nothing displayed for this case.

web/src/beta/components/RangeSlider/index.tsx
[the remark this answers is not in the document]
nina992: @airslice  noted , thank you , fixed

web/src/beta/components/fields/Property/PropertyField/index.tsx
nina992: @airslice this is how it looks now is not right??
image

PR #825, opened 2023-11-20
chore(web): remember the editor settings

mkumbobeaty:
# Overview

## What I've done

- remember width resize


web/src/beta/components/Resizable/hooks.ts
```diff
@@ -142,9 +180,28 @@ export default (
...
+      localStorageKey,
+      JSON.stringify({
+        [localStorageKey]: {
+          size: initialSize,
+          minimized: false,
+          difference: 0,
+          startingSize: initialSize,
+        },
+      }),
+    );
```
KaWaite, line 194:
```suggestion
    if (localStorageKey) {
       localStorage.setItem(
         localStorageKey,
         JSON.stringify({
           [localStorageKey]: {
             size: initialSize,
             minimized: false,
             difference: 0,
             startingSize: initialSize,
           },
         }),
    )};
```

KaWaite: Seems like parsedStated isn't needed in this callback, right?
Also, I think we should allow general functionality of Resizable to still work even if there is an issue/bug around the local storage. Thats why you should wrap the localStorage use, rather than returning early and not allowing anything to happen.

web/src/beta/components/Resizable/index.tsx
```diff
@@ -21,15 +24,27 @@ const Resizable: React.FC<Props> = ({
...
   );
 
+  useEffect(() => {
+    const setSizeBasedOnTab = () => {
+      if (tab === "map" || tab === "story" || tab === "widgets") {
+        localStorageKey && setSizeFromStorage(localStorageKey);
+      }
+    };
+    setSizeBasedOnTab();
+  }, [tab, setSizeFromStorage, localStorageKey]);
```
KaWaite, line 46: First, do we really need to check the tab? Or can we just `if (!!localStorageKey) setSizeFromStorage(localStorageKey)`?
Second, Resizable is a reusable component, so shouldn't have such specific logic here anyways.

Instead of using a useEffect and the usecallback (setSizeFromStorage), you could probably just use a useMemo, no? 🤔

web/src/beta/components/Resizable/hooks.ts
```diff
@@ -40,16 +39,54 @@ export default (
...
+          setSize(storedSize);
+          setMinimized(parsedState.minimized);
+        } else {
+          setSize(initialSize);
+          setMinimized(false);
+        }
+      }
+    };
+    setSizeBasedOnTab();
+  }, [setSizeFromStorage, localStorageKey, getLocalStorageParsedState, initialSize]);
```
KaWaite, line 88: setSizeFromStorage isn't necessary

web/src/beta/components/Resizable/hooks.ts
```diff
@@ -131,7 +176,16 @@ export default (
...
-  }, [bindEventListeners, isResizing, unbindEventListeners]);
+  }, [
+    bindEventListeners,
+    difference,
+    isResizing,
+    localStorageKey,
+    minimized,
+    size,
+    startingSize,
+    unbindEventListeners,
```
KaWaite, line 187: Same here, curious why you are adding so many deps. Any deps already attached to bindEventListeners or unbindEventListeners when changed will update those callbacks, which in turn will refire this useEffect.
If you add them here too, I am not 100% sure, but it might make this useEffect fire MANY times more than necessary.

web/src/beta/components/Resizable/hooks.ts
```diff
@@ -142,9 +196,28 @@ export default (
   );
 
   const handleResetSize = useCallback(() => {
+    if (!localStorageKey) return;
```
KaWaite, line 199: Same as my previous comment from last week. I think we should allow basic functionality even if there is issues with localstorage

web/src/beta/features/Editor/index.tsx
```diff
@@ -224,21 +216,14 @@ const Editor: React.FC<Props> = ({ sceneId, projectId, workspaceId, tab }) => {
...
-          {rightPanel && (
-            <Resizable
-              direction="vertical"
-              gutter="start"
-              initialSize={metrics.propertyMenuWidth}
-              minSize={metrics.propertyMenuMinWidth}>
-              {rightPanel}
-            </Resizable>
-          )}
+          {rightPanel && rightPanel}
```
KaWaite, line 226: Just do {rightPanel} and {leftPanel} in this file, and then in the two panel's just return null instead of undefined and it should be fine 👍🏻

web/src/beta/features/Editor/useLeftPanel.tsx
```diff
@@ -73,32 +75,46 @@ export default ({
...
-            selectedSceneSetting={selectedSceneSetting}
-            onLayerDelete={onLayerDelete}
-            onLayerNameUpdate={onLayerNameUpdate}
-            onLayerVisibilityUpate={onLayerVisibilityUpate}
-            onLayerSelect={onLayerSelect}
-            onSceneSettingSelect={onSceneSettingSelect}
-            onDataSourceManagerOpen={onDataSourceManagerOpen}
-            onFlyTo={onFlyTo}
-          />
+          <Resizable
```
KaWaite, line 78: Hmm can you find a way not to have to duplicate Resizable this much? If we add more tabs in the future super quick this file would get much bigger than necessary.
Some options might be:
- Move in to SidePanel which is used by MapSidePanel and StorySidePanel as SidePanelCommon
- do a bit more in this leftPanel memo
- Move back to where it was (for now)
- etc

web/src/beta/components/Resizable/hooks.ts
```diff
@@ -104,7 +142,14 @@ export default (
     setPosition({ x: 0, y: 0 });
     setStartingSize(size);
     setDifference(0);
-  }, [isResizing, size]);
+    if (!localStorageKey) return;
```
KaWaite, line 145: You actually don't need any of these returns I don't think since you are returning inside the `setLocalStorageData`

web/src/beta/features/Editor/useLeftPanel.tsx
mkumbobeaty: @KaWaite  Think we need to discuss this

web/src/beta/components/SidePanelSectionField/index.tsx
```diff
@@ -11,13 +11,17 @@ const SidePanelSectionField: React.FC<{
...
+  const [opened, setOpened] = useState<boolean>(() => {
+    const storedValue = localStorage.getItem(storageKey || "");
+    return storedValue ? JSON.parse(storedValue) : !startCollapsed ?? true;
+  });
 
   useEffect(() => {
-    setOpened(!startCollapsed ?? true);
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+    localStorage.setItem(storageKey || "", JSON.stringify(opened));
+  }, [opened, storageKey]);
```
KaWaite, line 24: opened isn't needed here, is it?

KaWaite (approved):

airslice (approved):

PR #831, opened 2023-11-22
chore: fix title's panel setting

nina992:
# Overview
Before : 
image

After:
image

### DIFF
--- a/web/src/beta/lib/core/StoryPanel/Block/builtin/Title/index.tsx
+++ b/web/src/beta/lib/core/StoryPanel/Block/builtin/Title/index.tsx
@@ -14,7 +14,12 @@ export type Props = BlockProps;
 const TitleBlock: React.FC<Props> = ({ block, isSelected, ...props }) => {
   const t = useT();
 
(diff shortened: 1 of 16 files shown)


web/src/beta/lib/core/StoryPanel/Block/builtin/Title/index.tsx
```diff
@@ -26,14 +26,20 @@ const TitleBlock: React.FC<Props> = ({ block, isSelected, ...props }) => {
     [property?.title?.color?.value],
   );
   const hasEmptySpace = isEmptyString(title);
+  const titleProperty = useMemo(() => {
+    return {
+      ...property,
+      panel: { padding: property?.title?.padding },
+    };
+  }, [property]);
```
KaWaite, line 34: Can you do this up a level?
You are already pulling out title from the page and then pass that title property in to this Title "story block". If you process fully where you pull the title group out of page we just have one place to look for issues in processing in the future or adding new fields/groups/etc

KaWaite (approved):

PR #839, opened 2023-11-28
fix(web): support built-in widget visible on WAS

airslice:
# Overview

`Visible` field has been added to built-in widgets since [this](https://github.com/reearth/reearth/pull/380) PR.
But only NLS WAS supports that feature. We need to support the same thing without NLS since the widget manifest is shared and from user view this is a bug that not working.

## What I've done
- Update classic classic WAS to support visible feature for built-in widgets.
- Refactor some on visible implement on classic / classic NLS / beta.
  - Calculate `invisibleWidgetIDs` in higher level. Layout (isMobile) / alignSystem / Builtin widget visible property change will trigger re-calculate.
  - Move invisible check to higher level (in Areas). There will be empty container (will effect gap) if check in widget component level.


web/src/beta/lib/core/Crust/Widgets/useWidgetAlignSystem.ts
```diff
@@ -1,15 +1,27 @@
...
+  type WidgetLocationOptions,
+  type WidgetArea,
+  type WidgetSection,
+  type WidgetZone,
 } from ".";
 
-export default ({ alignSystem }: { alignSystem: WidgetAlignSystem | undefined }) => {
+const sections: (keyof WidgetZone)[] = ["left", "center", "right"];
+const areas: (keyof WidgetSection)[] = ["top", "middle", "bottom"];
+const zones: (keyof WidgetAlignSystem)[] = ["outer", "inner"];
```
KaWaite, line 16: As a rule (both in core and in reearth), can we start putting constants in a `constants.ts` file, uppercasing the names with underscores (ie. SECTIONS, DEFAULT_BACKGROUND_COLOR, etc) 🙏🏻

web/src/beta/lib/core/Crust/Widgets/useWidgetAlignSystem.ts
```diff
@@ -18,23 +30,30 @@ export default ({ alignSystem }: { alignSystem: WidgetAlignSystem | undefined })
   //       The reason why we use invisible list is prevent initializing cost.
```
KaWaite, line 30: Can you update this comment for the invisible widget IDs to be more understandable?

web/src/beta/lib/core/Crust/Widgets/Widget/builtin/Navigator/hooks.ts
```diff
@@ -26,16 +25,14 @@ export default function ({
...
-  onVisibilityChange?: (id: string, visible: boolean) => void;
+  onVisibilityChange?: () => void;
 }) {
   const [degree, setDegree] = useState(0);
   const [isHelpOpened, setIsHelpOpened] = useState(false);
   const orbitRadianRef = useRef(0);
   const isMovingOrbit = useRef(false);
-  const visible = useVisible({
-    widgetId: widget.id,
+  useVisible({
```
KaWaite, line 34: Not quite understanding why we need these useVisible functions being called inside of the widgets when we are checking, and filtering out, any invisible widgets in WAS's Area component.
Can you explain that here or on slack? 🙏🏻

airslice: Removed. I think it's unnecessary now.

web/src/beta/lib/core/Crust/Widgets/useWidgetAlignSystem.ts
airslice: I kept this comment as-is from previous code...

web/src/beta/lib/core/Crust/Widgets/useWidgetAlignSystem.ts
airslice: Updated!

KaWaite (approved):

web/src/classic/components/molecules/Visualizer/WidgetAlignSystem/MobileZone.tsx
```diff
@@ -40,16 +41,19 @@ export default function MobileZone({
...
+  invisibleWidgetIDs,
   onWidgetAlignAreaSelect,
   children,
   ...props
 }: Props) {
   const filteredSections = useMemo(() => {
     return sections.filter(
       s =>
-        areas.filter(a => zone?.[s]?.[a]?.widgets?.length).length || (s === "center" && children),
+        areas.filter(a => zone?.[s]?.[a]?.widgets?.find(w => !invisibleWidgetIDs?.includes(w.id)))
```
keiya01, line 52: IMO: `some` is better, but `find` is also ok.
```suggestion
        areas.filter(a => zone?.[s]?.[a]?.widgets?.some(w => !invisibleWidgetIDs?.includes(w.id)))
```

web/src/classic/components/molecules/Visualizer/WidgetAlignSystem/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+const areas = ["top", "middle", "bottom"] as const;
+
+export const filterSections = (
+  zone?: WidgetZone,
+  invisibleWidgetIDs?: string[],
+  cb?: (s: (typeof sections)[number]) => void,
+) => {
+  return sections.filter(
+    s =>
+      areas.filter(a => zone?.[s]?.[a]?.widgets?.find(w => !invisibleWidgetIDs?.includes(w.id)))
```
keiya01, line 15: Same with [this](https://github.com/reearth/reearth/pull/839/files#r1408956105)

keiya01 (approved):

web/src/classic/components/molecules/Visualizer/WidgetAlignSystem/MobileZone.tsx
airslice: Updated. Thank you!

web/src/classic/components/molecules/Visualizer/WidgetAlignSystem/utils.ts
airslice: Updated. Thank you!

PR #853, opened 2023-12-05
chore(web): timeline with padding

mkumbobeaty:
# Overview

## What I've done

- Fix on click for the timeline block after page padding set
- Fix UI for timeline block after block padding


web/src/beta/lib/core/StoryPanel/Block/builtin/Timeline/hook.ts
```diff
@@ -45,6 +51,7 @@ type TimelineProps = {
 };
 
 type TimeEventHandler = (time: Date, committerId: string) => void;
+const DEFAULT_PADDING = 40;
```
KaWaite, line 54: You have DEFAULT_PADDING declared in multiple places.
When this happens, please declare it and export it from the lowest level and import it with the component/hook being imported above.

web/src/beta/lib/core/StoryPanel/Block/builtin/Timeline/hook.ts
```diff
@@ -101,6 +110,30 @@ export default ({
...
+      padding?.right || 0,
+      panelSettings?.padding.value.left || 0,
+      panelSettings?.padding.value.right || 0,
+    );
+  }, [
+    padding?.left,
+    padding?.right,
+    panelSettings?.padding.value.left,
+    panelSettings?.padding.value.right,
+  ]);
```
KaWaite, line 135: What I undestand is happening here is that you are cross-referencing the storypanel's and the block's left/right padding and using whichever is highest for the block's padding.
Is that correct? If its correct, that isn't how it should be, right? Storypanel should use the storypanel's padding, and each block should use their own padding...
Maybe I'm misunderstanding something 🤔

mkumbobeaty: @KaWaite Yes you are right but both of padding (storypanel's padding, and  block padding) are needed. for example if padding is applied in a block I use `panelSettings?.padding.value.left || 0, panelSettings?.padding.value.right || 0,` to check for the specific block padding. and the remaining timeline block are not affected by this padding value

web/src/beta/lib/core/StoryPanel/Block/builtin/Timeline/hook.ts
```diff
@@ -337,22 +351,49 @@ export default ({
...
   ]);
+  const blockRef = useRef<HTMLDivElement>(null);
+
+  const handleResize = useCallback(() => {
+    if ((padding || panelSettings?.padding.value) && blockRef.current) {
+      const blockWidth = blockRef.current.offsetWidth;
+      const thresholdWidth = 360;
+      setIsMinimized(blockWidth < thresholdWidth);
+    }
+  }, [padding, panelSettings?.padding.value]);
```
KaWaite, line 362: We don't need padding or panelSettings anymore here, right?

KaWaite (approved):

PR #860, opened 2023-12-15
chore(web): add transition appearance

keiya01:
# Overview

I added an API to enable moving entity smoothly.

https://github.com/reearth/reearth/assets/34934510/a65d3c10-ac19-4bc4-8b26-56e7fd463863

## What I've done
- Added transition appearance
- Added support transition appearance's translate property in marker.

## What I haven't done
- Support transition for other feature type.
- Support transition appearance other than translate property.

## How I tested

You can use this script.(If you move the cursor to out of globe, it causes the exception error. I will make an API to prevent it in another PR)
```js
const id = reearth.layers.add({
  type: "simple",
  data: {
    type: "geojson",
    value: {
      type: "Feature",
      geometry: {
        type: "Point",
        coordinates: [0, 0, 100],
      },
    },
  },
  marker: {
    color: "#00FFFF",
  },
});

setTimeout(() => {
  reearth.camera.flyTo(id);
  reearth.on("mousemove", ({ lng, lat, height }) => {
    reearth.layers.override(id, {
      type: "simple",
      data: {
        type: "geojson",
        value: {
          type: "Feature",
          geometry: {
            type: "Point",
            coordinates: [0, 0, 100],
          },
        },
      },
      marker: {
        color: "#00FFFF",
      },
      transition: {
        useTransition: true,
        translate: [lng, lat, height],
      },
    });
  });
}, 10);
```


airslice (approved):

web/src/beta/lib/core/engines/Cesium/Feature/index.tsx
```diff
@@ -26,7 +26,16 @@ export * from "./utils";
 export { context, type Context } from "./context";
 export { getTag } from "./utils";
 
-const components: Record<keyof AppearanceTypes, [FeatureComponent, FeatureComponentConfig]> = {
+const NON_RENDERABLE_APPEARANCE = ["transition"] as const;
+const isRenderableAppearance = (
+  k: keyof AppearanceTypes,
+): k is Exclude<keyof AppearanceTypes, (typeof NON_RENDERABLE_APPEARANCE)[number]> =>
+  !(NON_RENDERABLE_APPEARANCE as unknown as string[]).includes(k);
```
pyshx, line 33: The casting (NON_RENDERABLE_APPEARANCE as unknown as string[]) is a bit awkward. It's necessary due to the as const assertion, but it might be clearer to define NON_RENDERABLE_APPEARANCE without as const if the tuple does not need to be immutable.

pyshx (approved):

web/src/beta/lib/core/engines/Cesium/Feature/index.tsx
keiya01: I want to keep this being type safe, so I fixed it like this. How do you think?:pray:
https://github.com/reearth/reearth/pull/860/commits/dd1c428b46ac7b6dcb1c9707369b3f37f981f591

pyshx: Looks Neat, Thanks.

PR #872, opened 2024-01-16
chore(web): basic sketch system

airslice:
# Overview

This PR adds the basic support for sketch system.

Features:
- Common:
  - Supported sketch types: `marker` `polyline` `circle` `rectangle` `polygon` `extrudedCircle` `extrudedRectangle` `extrudedPolygon`.
  - Set sketch type to undefined will exit sketch.
  - New created sketch feature will be added to selected sketch layer.
  - New created sketch feature will be auto selected and apply selectedFeatureColor.
  - Entities displayed while drawing and after drawing are different.
    - While drawing: Skech features -> color can be set with plugin API.
    - After drawing: Rendered as NLS feature -> apply layer appearance.
  - Support keypress `Esc` to cancel.
  - Support keypress `Space` to switch to move mode while drawing.

- In Editor:
  - Sketch tool will be avaliable only when a sketch layer been selected. Aka you need to create a sketch layer first.

- With plugin API:
  - You can add sketch feature to selected sketch layer (like in editor).
  - You can also start drawing directly, a new sketch layer will be added when finish drawing. 
    - In this case you can set the default appearance for this newly created layer with plugin API.

- Plugin API:
```typescript
// already exist:
reearth.interactionMode.override: (mode: InteractionModeType) => void;

// new added:
reearth.sketch = {
  readonly setType?: (type: SketchType) => void;
  // This is the color for dynamic entities displayed while drawing.
  readonly setColor?: (color: string) => void;
  readonly setDefaultAppearance?: (appearance: SketchAppearance) => void;
}

reearth.on('sketchfeaturecreated', ({
  layerId: string;
  featureId: string;
}) => void);

reearth.on('sketchtypechange', (type: SketchType | undefined) => void);
``` 

## What I've done

- Add Toolbar to reearth Editor.
- Add Sketch entity to reearth core.
- Add Sketch logic to reearth Map.
- Add plugin API.
- Add `selectedFeatureColor` to marker, polygon and polyline.
- Refactor mouse events.
- Refactor some on feature Polygon extrudedHeight property.

## What I haven't done

- Custom properties of sketch feature.
- Default style for sketch layers.
  - Should be created when create a sketch layer.
- Re-edit / delete sketch feature.
  - Add property panel on right panel. 
- Use backend feature API.

## Known issue:
  - Select clamped polygon [may have some bug](https://community.cesium.com/t/poor-pick-result-with-polgons/20853/8).

## How I tested

Note:
- Remember to turn on `Hide objects under terrian` if you enabled terrian.

- Default sketch layer appearance: 
(should be created as a new style and auto selected when create a new sketch layer, but this havn't been implemented yet so we need to manually create this style and link it to the sketch layer)

``` json
{
  "marker": {
    "height": 0,
    "heightReference": "clamp",
    "hideIndicator": true,
    "selectedFeatureColor": "#00bebe"
  },
  "polygon": {
    "classificationType": "terrain",
    "extrudedHeight": {
      "expression": "${extrudedHeight}"
    },
    "fillColor": "#FFFFFF",
    "height": 0,
    "heightReference": "clamp",
    "hideIndicator": true,
    "selectedFeatureColor": "#00bebe",
    "shadows": "enabled"
  },
  "polyline": {
    "clampToGround": true,
    "height": 0,
    "hideIndicator": true,
    "selectedFeatureColor": "#00bebe",
    "shadows": "enabled",
    "strokeColor": "#FFFFFF",
    "strokeWidth": 2
  }
}
```

Test plugin:
[sketch.zip](https://github.com/reearth/reearth/files/14091933/sketch.zip)


web/src/beta/lib/core/engines/Cesium/Sketch/utils.ts
```diff
@@ -0,0 +1,72 @@
+import { Cartesian3, type PolygonHierarchy, type Scene } from "@cesium/engine";
```
keiya01, line 1: Some functions are defined in [here](https://github.com/reearth/reearth/blob/main/web/src/beta/lib/core/engines/Cesium/Feature/HeatMap/utils.ts) as well, so could you update it to define in common place like `Cesium/utils/polygon.ts` or something?

web/src/beta/lib/core/engines/Cesium/Sketch/SurfaceMeasurement.tsx
```diff
@@ -0,0 +1,79 @@
...
+        }),
+        classificationType: ClassificationType.TERRAIN,
+        clampToGround: true,
+      },
+    }),
+    [color, positionsProperty],
+  );
+
+  const visualizer = useVisualizer();
+  visualizer.current?.engine.requestRender();
```
keiya01, line 48: We should use `useContext` from [here](https://github.com/reearth/reearth/blob/3bb0a22483a8c75a80791ca5c434ac9ec15fd093/web/src/beta/lib/core/engines/Cesium/Feature/context.ts).
We have to think about relationship for each modules, but we should use the context for now. `useVisualizer` is for user of NLS.

web/src/beta/lib/core/engines/Cesium/Sketch/ScreenSpaceElement.tsx
```diff
@@ -0,0 +1,96 @@
...
+      const xyz = visualizer.current?.engine.toXYZ(camera?.lng, camera?.lat, camera?.height);
+      if (!xyz) return;
+      occluder.cameraPosition = new Cartesian3(xyz[0], xyz[1], xyz[2]);
+      if (
+        windowPosition == null ||
+        windowPosition.x < 0 ||
+        windowPosition.y < 0 ||
+        windowPosition.x > window.innerWidth ||
+        windowPosition.y > window.innerHeight ||
+        !occluder.isPointVisible(position)
```
keiya01, line 66: We might be able to use [this function](https://github.com/reearth/reearth/blob/3bb0a22483a8c75a80791ca5c434ac9ec15fd093/web/src/beta/lib/core/engines/Cesium/useEngineRef.ts#L217).

web/src/beta/lib/core/engines/Cesium/Sketch/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    const color = useMemo(
+      () => Color.fromCssColorString(stringColor ?? DEFAULT_SKETCH_COLOR),
+      [stringColor],
+    );
+
+    return (
+      <>
+        {positionsArray?.map((positions, index) => (
+          <PolylineEntity key={index} dynamic positions={positions} color={color} />
```
keiya01, line 75: This is just an idea or a memo or something like that...
You are using an Entity for polygon and polyline directly, but we already have Polygon and Polyline feature and appearance. I think we might want to customize the appearance for each sketched feature in the future.
However we need to implement each appearance to each feature. For example, we already have the appearance for Polygon feature, resource feature and sketch feature. Managing these three feature one by one is confusing us, so we should make a function which manage each appearance in one place later.

web/src/beta/lib/core/engines/Cesium/Sketch/ExtrudedPolygonEntity.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const extrudedHeightProperty = useConstant(
+    () => new CallbackProperty(() => extrudedHeightRef.current, false),
+  );
+  const extrudedHeight = dynamic ? extrudedHeightProperty : extrudedHeightProp;
+
+  // Non-constant callback property in color doesn't request render in every
+  // frame; prevents it from flashing when the color changes instead.
+  const colorRef = useRef(color);
+  colorRef.current = color;
+  const colorProperty = useConstant(() => new CallbackProperty(() => colorRef.current, false));
```
keiya01, line 53: Using `CallbackProperty` forces disabling the request render mode. So CallbackProperty should be used only when sketching layer.

web/src/beta/lib/core/engines/Cesium/Sketch/Entity.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { forwardRef, memo, useEffect, useState } from "react";
+import { useCesium } from "resium";
+
+import { assignForwardedRef } from "./utils";
+
+export interface EntityProps extends CesiumEntity.ConstructorOptions {
+  entities?: EntityCollection;
+}
+
+export const Entity = memo(
```
keiya01, line 12: We already have shared Entity component in [here](https://github.com/reearth/reearth/blob/3bb0a22483a8c75a80791ca5c434ac9ec15fd093/web/src/beta/lib/core/engines/Cesium/Feature/utils.tsx#L74). Can't we use this? This attaches some tags to identify the feature, so selectedFeature should work by this component.

web/src/beta/lib/core/engines/Cesium/Feature/Polyline/index.tsx
```diff
@@ -50,7 +52,18 @@ export default function Polyline({ id, isVisible, property, geometry, layer, fea
     [coordinates ?? []],
     isEqual,
   );
-  const material = useMemo(() => toColor(strokeColor), [strokeColor]);
+
+  const { viewer } = useCesium();
+  const entity = viewer ? findEntity(viewer, layer?.id, feature?.id) : undefined;
```
keiya01, line 57:
```suggestion
  const entity = useMemo(() => viewer ? findEntity(viewer, layer?.id, feature?.id) : undefined, []);
```
This process should be wrapped by useMemo, because this function runs too heavy process.

keiya01, line 57: I think you can use the `ref` from [EntityExt](https://github.com/reearth/reearth/blob/a9d933fdd8782facef7393dac4a6522bc980360d/web/src/beta/lib/core/engines/Cesium/Feature/Polyline/index.tsx#L79) to get the entity without searching all entities.

web/src/beta/lib/core/engines/Cesium/Sketch/Entity.tsx
keiya01: Ah I understood. The sketched entity is added as a general feature after an Entity is sketched, right? Then this is no problem, that's make sense :pray:

web/src/beta/lib/core/engines/Cesium/Sketch/utils.ts
airslice: Updated. Merged into `Cesium/utils/polygon.ts`.

web/src/beta/lib/core/engines/Cesium/Sketch/ScreenSpaceElement.tsx
airslice: Updated with context. Thank you!

web/src/beta/lib/core/engines/Cesium/Sketch/ExtrudedPolygonEntity.tsx
airslice: Sketch entities will only be rendered while drawing so i think it's okey. BTW i removed the callbackProperty for color since it will not change while drawing so we can use color directly.

web/src/beta/lib/core/engines/Cesium/Sketch/Entity.tsx
airslice: Yes the sketch entities will be used only when drawing. After one feature been created it will be rendered with NLS Feature instead.

web/src/beta/lib/core/engines/Cesium/Feature/Polyline/index.tsx
airslice: Updated with ref. Thank you!

keiya01 (approved):

PR #876, opened 2024-01-22
chore(web): timeline bug fixing

mkumbobeaty:
# Overview

## What I've done

- Improved timezone in timeline block based on what is set.
Screenshot from 2024-01-22 09-35-00
Screenshot from 2024-01-22 09-35-24

- Improve timeline settings for the input to have same timezone if not should not be able to apply settings

Screenshot from 2024-01-22 09-34-18

-Add two speeds in drop down (1se/sec and 0.5min/sec)

Screenshot from 2024-01-22 09-40-30


web/src/beta/components/fields/TimelineField/EditPanel/index.tsx
```diff
@@ -32,6 +35,18 @@ const EditPanel = ({
     setTimelineValues,
   });
 
+  const checkTimezone = useMemo(() => {
```
airslice, line 38: Maybe naming directly like `timezoneMatches` will be better?

airslice (approved):

web/src/beta/components/fields/TimelineField/EditPanel/index.tsx
airslice: Updated.

PR #880, opened 2024-01-26
chore(web): infobox support in beta

KaWaite:
# Overview

## general
- Update codeowners

## web
- Refactor storytelling panel code in to reusable components to be shared between it and the infobox
- Replace old infobox implementation (which wasn't being used in beta) with beta infobox
- Add 3 builtin blocks (image, text and property)
- Add plugin infobox block support


pyshx (approved):
`web/src/services` and `server/` changes Approved

PR #883, opened 2024-01-31
chore(web): add rotateCameraOnCenter

mkumbobeaty:
# Overview

## What I've done
- Added rotate camera on center API


keiya01 (approved):

web/src/beta/lib/core/Crust/Plugins/api.ts
```diff
@@ -540,6 +542,7 @@ export function commonReearth({
         flyToGround,
         setView,
         flyToBBox,
+        rotateCameraOnCenter,
```
airslice, line 545: Since it's already a method of `camera`, naming as `rotateOnCenter` will be better?

PR #884, opened 2024-02-01
chore(web): fix statement for Hideindicator and viewer.selectedEntity

m-abe-dev:
# Overview
deleted if statement for viewer.selectedEntity = undefined;

## Memo

First viewer.selectedEntity = undefined; (L541): This line is used within the handleClick function to clear the selection under certain conditions, such as when a specific type of entity is clicked but deemed inappropriate to remain selected. It clears the currently selected entity if the user's click action is determined not to result in a valid selection

Second viewer.selectedEntity = undefined; (L691): This line is placed at the end of the handleClick function and is used to ensure there is no selected entity left after the click event has been fully processed. It acts as a final check to reset the selection state if no selection should be made based on the user's action or specific conditions dictate that the selection should be cleared

### DIFF
--- a/web/src/beta/lib/core/engines/Cesium/hooks.ts
+++ b/web/src/beta/lib/core/engines/Cesium/hooks.ts
@@ -361,6 +361,8 @@ export default ({
     const tag = getTag(entity);
     if (entity instanceof Entity && !tag?.hideIndicator) {
       viewer.selectedEntity = entity;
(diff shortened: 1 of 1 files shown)


m-abe-dev:
close this PR because tag?.hideIndicator was fixed by new commit. 

!tag?.hideIndicator → tag?.hideIndicator

PR #899, opened 2024-02-21
chore(web): added plugins for camera rotate

mkumbobeaty:
# Overview

- Implemented `keyboardCameraRotate` plugin API that used on camera rotation when keyboard is enabled
- Added the `negatuvePiToPi ` plugin that export the ` CesiumMath.negatuvePiToPi `
`  window.reearth?.scene?.negatuvePiToPi(cameraPotion?.roll)
`


web/src/beta/lib/core/engines/Cesium/useEngineRef.ts
```diff
@@ -167,6 +167,12 @@ export default function useEngineRef(
           cart.height,
         ];
       },
+
+      negatuvePiToPi: (angle: number) => {
+        const viewer = cesium.current?.cesiumElement;
+        if (!viewer || viewer.isDestroyed()) return;
+        return Math.abs(CesiumMath.negativePiToPi(angle)) > Math.PI / 86400;
+      },
```
airslice, line 175: The return value seems to be boolean which does not match the naming of the function. If this function is supposed to be a simple export of CesiumMath.negativePiToPi then just return it and handle the other logic at where it been used. Otherwise please rename with other proper names.

airslice, line 175: The return value seems to be boolean which does not match the naming of the function. If this function is supposed to be a simple export of CesiumMath.negativePiToPi then just return it and handle the other logic at where it been used. Otherwise please rename with other proper name.

web/src/beta/lib/core/engines/Cesium/useEngineRef.ts
```diff
@@ -468,6 +474,19 @@ export default function useEngineRef(
         const assignments = getOverriddenScreenSpaceCameraOptions(options);
         Object.assign(controller, assignments);
       },
+
+      keyboardCameraRotate: () => {
```
airslice, line 478: The word `keyboard` seems meanning less here. I guess it's from the behavior of View 3.0, but as a API of reearth it should be something like `rollCameraHorizontal`

web/src/beta/lib/core/Map/types/index.ts
```diff
@@ -100,6 +100,7 @@ export type EngineRef = {
     z: number,
     options?: { useGlobeEllipsoid?: boolean },
   ) => [lng: number, lat: number, height: number] | undefined;
+  negatuvePiToPi: (angle: number) => void;
```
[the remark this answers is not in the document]
airslice: BTW i just checked Cesium negatuvePiToPi and seems not complex, is it necessary to provide this math function in plugin API?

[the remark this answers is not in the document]
keiya01: If possible, we should copy and paste the function in VIEW3.

web/src/beta/lib/core/engines/Cesium/useEngineRef.ts
keiya01: We can do this just by using `reearth.camera.setView` in plugin side. So I think we don't need this.

mkumbobeaty: @keiya01 If we use reearth.camera.setView for the plugin we will have to pass the camera position which delays as @airslice mentioned last time and make the camera not smooth

keiya01: I think `reearth.camera.setView` doesn't update some state in ReEarth. It just update the Cesium's camera:eyes:

PR #921, opened 2024-03-13
chore(web): support draw clipping for 3dtiles

airslice:
# Overview

This PR supports a new type of clipping on 3dtiles. It will generate side planes with a polygon and top/bottom plane with relative height (base height is the sampled terrain height of center point of the polygon).

## What I've done

- Add draw clipping to 3d tiles appearance `experimental_clipping.draw`
  - Basially i don't want to effect the original clipping box, all draw clipping needed is in this `draw` property, it will have higher priority campare with clipping box.
- Refactor some on sketch API to better fit the requirement of draw clipping.

NOTE: Concave polygon is not supported (from engine limit)


web/src/beta/lib/core/engines/Cesium/Feature/Tileset/DrawClippingPolygon/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+    () =>
+      coordiantes
+        ? new PolygonHierarchy(coordiantes.map(c => Cartesian3.fromDegrees(c[0], c[1], c[2] ?? 0)))
+        : undefined,
+    [coordiantes ?? []],
+    isEqual,
+  );
+
+  const strokeColor = useMemo(() => toColor("#00bebe"), []);
+  const fillColor = useMemo(() => toColor("#00bebe11"), []);
```
keiya01, line 35: Should these color be customizable?

web/src/beta/lib/core/engines/Cesium/Sketch/ExtrudedPolygonEntity.tsx
```diff
@@ -53,12 +55,18 @@ export const ExtrudedPolygonEntity: FC<ExtrudedPolygonEntityProps> = ({
         extrudedHeight,
         extrudedHeightReference: HeightReference.RELATIVE_TO_GROUND,
         fill: true,
+        outline: true,
+        outlineWidth: 1,
```
keiya01, line 59: Should the outline be customizable from plugin API?

keiya01 (approved):

web/src/beta/lib/core/engines/Cesium/Feature/Tileset/DrawClippingPolygon/index.tsx
airslice: Yes better have that. I updated the style support in appearance.

web/src/beta/lib/core/engines/Cesium/Sketch/ExtrudedPolygonEntity.tsx
airslice: For the sketch API i want to have a refactor in another PR. So i'll skip this one here.

PR #937, opened 2024-03-23
chore(web): sketch layer step3

mkumbobeaty:
# Overview

- Refactor the initial implementation for custom properties schema

- Added custom fields based on custom properties schema  to update the feature properties of the sketch layer


pyshx (approved):
`services` and `beta/../visualizer` changes looks good to me.

web/src/beta/lib/core/Map/Sketch/hooks.ts
```diff
@@ -236,6 +236,7 @@ export default function useHooks({
             : pluginSketchLayerFeatureAdd(selectedLayer, feature);
 
         if (layerId && featureId) {
+          console.log("callled");
```
airslice, line 239:
```suggestion

```

web/src/beta/features/Editor/SketchLayerManager/index.tsx
```diff
@@ -55,20 +57,26 @@ const SketchLayerManager: React.FC<SketchProps> = ({ sceneId, layerStyles, onClo
   }, []);
 
   const handleSubmit = () => {
+    const schemaJSON = propertyList.reduce((acc, property, index) => {
+      const [key] = Object.keys(property);
+      const value = `${property[key]}_${index + 1}`;
```
airslice, line 62: Why do we need the value to be like `Int_1`, is this requirement from backend?

airslice: Oh it's for the sort? Is it necessary?

web/src/beta/features/Editor/tabs/map/RightPanel/LayerInspector/FeatureData.tsx
```diff
@@ -1,23 +1,86 @@
...
+      .sort((a, b) => {
+        const aIndex = parseInt(a.value.split("_")[1]);
+        const bIndex = parseInt(b.value.split("_")[1]);
+        return aIndex - bIndex;
+      });
+
+    const fieldArray = sortedValues.map(({ key, value }) => ({
+      id: uuidv4(),
+      type: value.replace(/_\d+$/, ""),
+      title: key.toLowerCase(),
```
airslice, line 64: Why do we need to use lower case? I think the title should present what we have in data.

web/src/beta/features/Editor/SketchLayerManager/index.tsx
airslice: Emmm maybe it's okey, but better add a comment like `// record order for sort later`, i feel it is a little bit difficult for other developers to understand on first read.

web/src/beta/features/Editor/tabs/map/RightPanel/LayerInspector/CustomPropertField.tsx
```diff
@@ -0,0 +1,105 @@
...
+      value={getDynamicValue(selectedFeature, field.title, field.value)}
+      onChange={handleChange}
+    />
+  ) : field?.type === "Boolean" ? (
+    <ToggleField
+      key={field?.id}
+      name={field?.title}
+      checked={getDynamicValue(selectedFeature, field.title, field.value)}
+      onChange={handleChange}
+    />
```
airslice, line 101: Why some fields has `key` while the others not?

web/src/beta/features/Editor/SketchLayerManager/index.tsx
mkumbobeaty: @airslice yes I will add the comment, this is for sorting cause when the custom properties send to backend they are ordered in alphabetic order, but what we want is user to see the custom properties in the order  he added them

web/src/beta/components/TabMenu/index.tsx
```diff
@@ -18,11 +18,18 @@ export type TabObject = {
 export type Props = {
   tabs: TabObject[];
   selectedTab: string;
+  isScroll?: boolean;
```
airslice, line 21: Could you replace all `isScroll` with `scrollable`? I think `isScroll` is a little bit confusing.

web/src/beta/features/Editor/SketchLayerManager/CustomedProperties/index.tsx
```diff
@@ -1,107 +1,70 @@
...
-        </PropertyContentWrapper>
-      </PropertyList>
-
+          )}
+        />
+      )}
+      <AddButtonWrapper>
+        <Button icon="plus" text={t("New Property")} size="small" onClick={handlePropertyAdd} />
+      </AddButtonWrapper>
       <SubmitWrapper />
```
airslice, line 68:
```suggestion
```
Seems doesn't in use any more?

airslice (approved):

PR #949, opened 2024-03-29
fix(web): load cesium token from separate file

airslice:
# Overview
This is base on [this PR](https://github.com/reearth/reearth/pull/907).

## What I've done

- Fix checking the fetch result with string length === 177
- Support fetching remote cesium ion token.
  - Add a new optional env var REEARTH_WEB_CESIUM_ION_TOKEN_URL


web/vite.config.ts
```diff
@@ -93,11 +93,18 @@ function config(): Plugin {
       const remoteReearthConfig = envs.REEARTH_WEB_CONFIG_URL
         ? await (await fetch(envs.REEARTH_WEB_CONFIG_URL)).json()
         : {};
+      const reqForRemoteCesiumIonToken = envs.REEARTH_WEB_CESIUM_ION_TOKEN_URL
+        ? await fetch(envs.REEARTH_WEB_CESIUM_ION_TOKEN_URL)
+        : undefined;
+      const remoteCesiumIonToken = reqForRemoteCesiumIonToken?.ok
+        ? await reqForRemoteCesiumIonToken.text()
+        : "";
```
KaWaite, line 101:
```suggestion
      const remoteCesiumIonToken = envs.REEARTH_WEB_CESIUM_ION_TOKEN_URL
        ? await (await fetch(envs.REEARTH_WEB_CESIUM_ION_TOKEN_URL)).text()
        : undefined;
```

I think if you fallback to an empty string and not undefined, the ternary operator below will always be true. So either this change or just remove the ternary operator below so we will always have cesiumIonAccessToken in config.

KaWaite (approved):

mkumbobeaty (approved):

pyshx (approved):

web/vite.config.ts
airslice: This part has been refacted.

PR #951, opened 2024-04-04
feat(web): update cesium

keiya01:
# Overview

I updated Cesium to use latest version of Cesium.

There are a lot of update.
- Some API like Model, 3DTiles, Terrain and ImageryLayer is loaded asynchronously.
- Rendering process seems to be changed a bit. We need to invoke requestRender in additional part in beta.

## What I've done

- Updated the code of classic, classic's NLS and beta.

## How I tested

- Build [this PR of resium](https://github.com/reearth/resium/pull/625).
- Replace `reearth/web/node_modules/resium/dist` by the generated `dist` directory of resium.
- Run `yarn start`.
- Test classic, classic's NLS(This might be unnecessary?) and beta(include VIEW3) entirely.


web/src/classic/components/molecules/Visualizer/Engine/Cesium/Tileset/index.tsx
```diff
@@ -232,13 +232,23 @@ const Tileset: FC<PrimitiveProps<Property, any, SceneProperty>> = memo(function
...
 
+  const handleOnReady = useCallback(
+    (t: Cesium3DTilesetType) => {
+      if (_debugFlight) {
+        viewer?.flyTo(t);
+      }
+      setIsTilesetReady(true);
+    },
+    [viewer],
+  );
```
KaWaite, line 243: Just to avoid any unnecessary re-renders because of unnecessary state updates.

```suggestion
  const handleOnReady = useCallback(
    (t: Cesium3DTilesetType) => {
      if (isTilesetReady) return;
      if (_debugFlight) {
        viewer?.flyTo(t);
      }
      setIsTilesetReady(true);
    },
    [viewer],
  );
```

web/src/beta/lib/core/engines/Cesium/useOverrideGlobeShader.ts
```diff
@@ -218,12 +218,15 @@ export const useOverrideGlobeShader = ({
     globe.vertexShadowDarkness = globeShadowDarkness ?? globe.vertexShadowDarkness;
   }, [cesium, globeShadowDarkness, hasVertexNormals]);
 
-  // This need to be invoked before Globe is updated.
-  useEffect(() => {
+  const originalGlobeFS = useRef<string>();
```
KaWaite, line 221: What is the point of this? You don't use it unless you are setting `GlobeFS`. But `GlobeFS`'s value comes either from this `originalGlobeFS` or `baseFragmentShaderSource....` (in that order). 
My understanding of this code is:
First render
1. `GlobeFS` is set to `baseFragmentSha.....`
2. `originalGlobeFS` is set to same
Second + render
1. `GlobeFS` is again set to originalGlobeFS
2. `originalGlobeFS` has a value, so skips.

This means that GlobeFS is ALWAYS going to be originalGlobeFS. I'm guessing you want GlobeFS to be `baseFragmentSh....`'s up-to-date value, no? 🤔

web/src/beta/lib/core/engines/Cesium/Feature/index.tsx
```diff
@@ -112,6 +113,17 @@ export default function Feature({
...
+  const { requestRender } = useContext();
+  // TODO: Find a way to wait updating the entity
+  useEffect(() => {
+    setTimeout(() => {
+      requestRender?.();
+    }, 300);
+  });
+
+  // Need to invoke requestRender when it's unmounted
+  useEffect(() => () => requestRender?.(), [requestRender]);
```
KaWaite, line 125: (optional) Can you add a brief reason why or at least a link to where its explained?

web/src/classic/components/molecules/Visualizer/Engine/Cesium/Tileset/index.tsx
keiya01: The event happens only once, also it's just boolean, so it doesn't cause unnecessary render.

web/src/beta/lib/core/engines/Cesium/Feature/index.tsx
keiya01: I couldn't the actual resource, but it's necessary...

web/src/beta/lib/core/engines/Cesium/useOverrideGlobeShader.ts
keiya01: Fixed! https://github.com/reearth/reearth/pull/951/commits/20ede6c3718a44e2a174a2d37ca1cacaaf572a88

airslice (approved):

mkumbobeaty (approved):

PR #967, opened 2024-04-23
chore(web): remove jotai provider for beta editor

airslice:
# Overview
Notification component is out of the jotai Provider which makes it can't be used in beta Editor.

## What I've done
Move the Provider to have the Notification included.
But somehow i feel the provider is unnecessary in this case.


airslice:
We have a better solution to refactor the status as local states instead of using Jotai. Closing this one.

PR #979, opened 2024-05-14
chore(web): refactor number input

mkumbobeaty:
# Overview
- Refactor number input
## What I've done
- handle float numbers


web/src/beta/lib/reearth-ui/components/NumberInput/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+    setIsFocused(true);
+  }, []);
+
+  // This function prevent defaut browser clear input on these keys pressed
+  // https://stackoverflow.com/questions/49869624/numeric-value-gets-clear-in-we-press-letter-e-in-input-type-number
+  const handleKeyDown = useCallback((e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === "e" || e.key === "+" || e.key === "-") {
+      e.preventDefault();
+    }
+  }, []);
```
airslice, line 62: This prevent user input negative number value as well. Maybe we need to think more for the solution.

web/src/beta/lib/reearth-ui/components/NumberInput/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const handleFocus = useCallback(() => {
+    setIsFocused(true);
+  }, []);
+
+  return (
+    <Wrapper size={size} status={isFocused ? "active" : "default"}>
+      <StyledInput
+        type="text"
+        inputMode="numeric"
+        pattern="[0-9]*"
```
airslice, line 58: Do we still need these since we already handle the validate with regexp?

mkumbobeaty: @airslice removed it

airslice (approved):

PR #988, opened 2024-05-23
chore(WIP): Support plugin block in Beta infobox

m-abe-dev:
# Overview

## What I've done
adding plugins query
スクリーンショット 2024-05-25 2 14 30

## What I haven't done
graphql mutation error
スクリーンショット 2024-05-25 2 15 37

## Memo

According to the error message, the addNLSInfoboxBlock mutation is designed to accept only extensions of type PluginExtensionType.Block. Therefore, attempting to add an InfoboxBlock type will result in an error. fixing Backend is needed to resolve this issue.


airslice:
We have already support beta's infobox plugin block with extension type `infoboxBlock`.

PR #993, opened 2024-05-27
chore(web): refactor date and time picker

mkumbobeaty:
# Overview

## What I've done

- Refactor date picker input 
- Refactor time picker input


web/src/beta/lib/reearth-ui/components/DatePicker/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  );
+
+  const handleBlur = useCallback(() => {
+    setIsFocused(false);
+    onBlur?.(currentValue);
+  }, [currentValue, onBlur]);
+
+  const handleFocus = useCallback(() => {
+    setIsFocused(true);
+  }, []);
```
ed0424, line 37: Do we need to wrap this function with useCallback?

mkumbobeaty: @ed0424 wrapping the function with useCallback helps in optimising performance and preventing unnecessary re-renders

web/src/beta/lib/reearth-ui/components/DatePicker/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  return (
+    <Wrapper status={isFocused ? "active" : "default"}>
+      <StyledInput
+        value={currentValue}
+        placeholder="MM/DD/YYYY"
```
airslice, line 43: I find it been displayed in lowercase in storybook, could you investigate it? I think it should display in uppercase as it defined here.

mkumbobeaty: @airslice since we are using native html date picker the provide their own placeholder which I can not override it. so I think this placeholder is not used may be I can remove it.

airslice: Emmm i think it's okey for now.

web/src/beta/lib/reearth-ui/components/DatePicker/index.tsx
airslice: Consider the use case here both using or not using this useCallback is acceptable to me. We may have a big refactor after react v19 release. We can keep this for now i think.

airslice (approved):

PR #997, opened 2024-05-29
chore(web): implement selector component

ed0424:
# Overview

### DIFF
--- a/web/src/beta/lib/reearth-ui/components/Button/index.tsx
+++ b/web/src/beta/lib/reearth-ui/components/Button/index.tsx
@@ -1,4 +1,4 @@
-import { FC } from "react";
+import { FC, MouseEvent } from "react";
 
 import { styled } from "@reearth/services/theme";
 
@@ -14,7 +14,7 @@ export type ButtonProps = {
   title?: string;
   extendWidth?: boolean;
   minWidth?: number;
-  onClick?: () => void;
+  onClick?: (e: MouseEvent<HTMLElement>) => void;
 };
 
 export const Button: FC<ButtonProps> = ({
--- a/web/src/beta/lib/reearth-ui/components/Button/index.tsx
+++ b/web/src/beta/lib/reearth-ui/components/Button/index.tsx
@@ -70,7 +70,7 @@ const StyledButton = styled("button")<{
   padding:
     size === "small"
       ? iconButton
-        ? `${theme.spacing.micro}px`
+        ? "2px"
         : `${theme.spacing.smallest}px ${theme.spacing.small}px`
       : iconButton
       ? `${theme.spacing.small}px`
@@ -103,5 +103,9 @@ const StyledButton = styled("button")<{
     color: `${theme.content.weaker}`,
     backgroundColor: appearance !== "simple" ? `${theme.bg[1]}` : "transparent",
     boxShadow: "none",
+  },
+  ["& svg"]: {
+    width: iconButton && size === "small" ? "12px" : "inherit",
+    height: iconButton && size === "small" ? "12px" : "inherit",
   },
 }));
--- a/web/src/beta/lib/reearth-ui/components/Button/index.tsx
+++ b/web/src/beta/lib/reearth-ui/components/Button/index.tsx
@@ -70,7 +70,7 @@ const StyledButton = styled("button")<{
   padding:
     size === "small"
       ? iconButton
-        ? "2px"
+        ? `${theme.spacing.micro}px`
         : `${theme.spacing.smallest}px ${theme.spacing.small}px`
       : iconButton
       ? `${theme.spacing.small}px`
--- a/web/src/beta/lib/reearth-ui/components/Selector/index.stories.tsx
+++ b/web/src/beta/lib/reearth-ui/components/Selector/index.stories.tsx
@@ -9,13 +9,38 @@ const meta: Meta<SelectorProps> = {
 export default meta;
 type Story = StoryObj<typeof Selector>;
 
-const LIST_ITEMS = ["item_1", "item_2", "item_3", "item_4", "item_5", "item_6"];
+const LIST_ITEMS = [
+  {
+    value: "item_1",
+    label: "item_1",
+  },
+  {
+    value: "item_2",
+    label: "item_2",
+  },
+  {
+    value: "item_3",
+    label: "item_3",
+  },
+  {
+    value: "item_4",
+    label: "item_4",
+  },
+  {
+    value: "item_5",
+    label: "item_5",
+  },
+  {
+    value: "item_6",
+    label: "item_6",
+  },
+];
 
 export const Default: Story = {
   render: () => {
     return (
       <div style={{ width: "300px" }}>
-        <Selector values={LIST_ITEMS} />
+        <Selector options={LIST_ITEMS} />
       </div>
     );
   },
@@ -25,7 +50,19 @@ export const MultipleSelector: Story = {
   render: () => {
     return (
       <div style={{ width: "300px" }}>
-        <Selector values={LIST_ITEMS} isMultiple />
+        <Selector options={LIST_ITEMS} multiple={true} />
+      </div>
+    );
+  },
+};
+
+export const Disabled: Story = {
+  render: () => {
+    return (
+      <div style={{ width: "300px" }}>
+        <Selector options={LIST_ITEMS} disabled />
+        <Selector options={LIST_ITEMS} value="item_1" disabled />
+        <Selector options={LIST_ITEMS} value={["item_1", "item_2"]} multiple disabled />
       </div>
     );
   },
--- a/web/src/beta/lib/reearth-ui/components/Selector/index.stories.tsx
+++ b/web/src/beta/lib/reearth-ui/components/Selector/index.stories.tsx
@@ -1,0 +1,32 @@
+import { Meta, StoryObj } from "@storybook/react";
+
+import { Selector, SelectorProps } from ".";
+
+const meta: Meta<SelectorProps> = {
+  component: Selector,
+};
+
+export default meta;
+type Story = StoryObj<typeof Selector>;
+
+const LIST_ITEMS = ["item_1", "item_2", "item_3", "item_4", "item_5", "item_6"];
+
+export const Default: Story = {
+  render: () => {
+    return (
+      <div style={{ width: "300px" }}>
+        <Selector values={LIST_ITEMS} />
+      </div>
+    );
+  },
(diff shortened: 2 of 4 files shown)


web/src/beta/lib/reearth-ui/components/Selector/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  const optionValues = useMemo(() => options, [options]);
+
+  useEffect(() => {
+    setSelectedValue(value ?? (multiple ? [] : undefined));
+  }, [value, multiple]);
+
+  useEffect(() => {
+    setSelectorWidth(selectorRef.current?.clientWidth);
+  }, [selectorRef]);
```
airslice, line 43: Maybe this is not enough. We may need a resizeObserver here, as you can imagin in reearth UI the side panel is resizeable, and the width could be changed anytime by user, we need to update this width once the selector width changes.

web/src/beta/lib/reearth-ui/components/Selector/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+  useEffect(() => {
+    if (selectorRef.current) {
+      const observer = new ResizeObserver(updateSelectorWidth);
+      observer.observe(selectorRef.current);
+      return () => {
+        observer.disconnect();
+      };
+    }
+    return () => {};
```
airslice, line 65:
```suggestion
```
Is there any special reason for why we need this empty return function?

web/src/beta/lib/reearth-ui/components/Selector/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+  }, [updateSelectorWidth]);
+
+  useEffect(() => {
+    if (selectorRef.current) {
+      const observer = new ResizeObserver(updateSelectorWidth);
+      observer.observe(selectorRef.current);
+      return () => {
+        observer.disconnect();
+      };
+    }
```
airslice, line 64: Is it possible to reuse one single ResizeObserver and do observe and disconnect when needed?

web/src/beta/lib/reearth-ui/components/Selector/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+  useEffect(() => {
+    setSelectedValue(value ?? (multiple ? [] : undefined));
+  }, [value, multiple]);
+
+  const updateSelectorWidth = useCallback(() => {
+    if (selectorRef.current) {
+      const newWidth = selectorRef.current.clientWidth;
+      setSelectorWidth(newWidth);
+      console.log("Selector width updated:", newWidth);
```
airslice, line 45:
```suggestion
```
Please remove console.log here.
To test the resize function you can add a story on storybook.

web/src/beta/lib/reearth-ui/components/Selector/index.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+export const Default: Story = {
+  render: () => {
+    return (
+      <div style={{ width: "300px" }}>
+        <Selector options={LIST_ITEMS} />
+      </div>
+    );
+  },
+};
+
```
airslice, line 48: Please add one more story to show the resize feature, you only need to remove the wrapper (div) so that the selector will cover all the width and the popup should has the same width when container width changes.

web/src/beta/lib/reearth-ui/components/Selector/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+    }
+  }, []);
+
+  useEffect(() => {
+    updateSelectorWidth();
+    window.addEventListener("resize", updateSelectorWidth);
+    return () => {
+      window.removeEventListener("resize", updateSelectorWidth);
+    };
```
airslice, line 54:
```suggestion
    return window.addEventListener("resize", updateSelectorWidth);
```
addEventListener will return a remove function, we can use it as return function of useEffect directly.

web/src/beta/lib/reearth-ui/components/Selector/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+  const renderTrigger = () => {
+    return (
+      <SelectInput isMultiple={multiple} isOpen={isOpen} disabled={disabled} width={selectorWidth}>
+        {!selectedValue?.length ? (
+          <Typography size="body" color={theme.content.weaker}>
+            {placeholder}
+          </Typography>
+        ) : multiple ? (
+          <SelectedItems>
```
airslice, line 117: Please doube comfirm with Red, currently multiple selected item will make the Selector higher, which i think should not happen (when it's still single line).

web/src/beta/lib/reearth-ui/components/Selector/index.tsx
m-abe-dev: i needed to fix this eslint error by adding empty return. 
This fix ensures that the useEffect hook using ResizeObserver always returns a cleanup function, making sure all code paths return a value. i thought this should resolve the TypeScript error.

スクリーンショット 2024-06-10 9 31 39

web/src/beta/lib/reearth-ui/components/Selector/index.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { Selector, SelectorProps } from ".";
+
+const meta: Meta<SelectorProps> = {
+  component: Selector,
+};
+
+export default meta;
+type Story = StoryObj<typeof Selector>;
+
+const LIST_ITEMS = [
```
mkumbobeaty, line 12: can we write this in small later

m-abe-dev: sure

web/src/beta/lib/reearth-ui/components/Selector/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+        />
+      </SelectInput>
+    );
+  };
+
+  return (
+    <SelectorWrapper ref={selectorRef}>
+      {disabled ? (
+        renderTrigger()
+      ) : (
```
airslice, line 164: Maybe we don't need this logic since Popup has a `disabled` prop?

web/src/beta/lib/reearth-ui/components/Selector/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+        <Popup
+          trigger={renderTrigger()}
+          open={isOpen}
+          onOpenChange={setIsOpen}
+          placement="bottom-start">
+          <DropDownWrapper width={selectorWidth}>
+            {optionValues.length === 0 ? (
+              <Typography size="body" color={theme.content.weaker}>
+                No option yet
+              </Typography>
```
airslice, line 174: Please keep the padding the same with normal option items.

web/src/beta/lib/reearth-ui/components/Button/index.tsx
```diff
@@ -67,7 +67,7 @@ const StyledButton = styled("button")<{
   padding:
     size === "small"
       ? iconButton
-        ? `${theme.spacing.micro}px`
+        ? "2px"
```
airslice, line 70: Why do we need to replace theme.spacing.micro with `2`? I think they're same value?

web/src/beta/lib/reearth-ui/components/Selector/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+    if (selectorRef.current) {
+      setSelectorWidth(selectorRef.current.clientWidth);
+    }
+  }, []);
+
+  useEffect(() => {
+    updateSelectorWidth();
+    const handleResize = () => updateSelectorWidth();
+    return window.addEventListener("resize", handleResize);
+  }, [updateSelectorWidth]);
```
airslice, line 63: I think we don't need this part since you already observe the selector element.
```suggestion
```

web/src/beta/lib/reearth-ui/components/Selector/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+    if (currentRef) {
+      (currentRef as any)._updateSelectorWidth = updateSelectorWidth;
+      resizeObserver.observe(currentRef);
+    }
+    return () => {
+      if (currentRef) {
+        resizeObserver.unobserve(currentRef);
+      }
+    };
+  }, [updateSelectorWidth]);
```
airslice, line 76:
```suggestion
  useEffect(() => {
    const selectorElement = selectorRef.current;
    if (!selectorElement) return;

    const resizeObserver = new ResizeObserver(entries => {
      if (!entries || entries.length === 0) return;
      const { width } = entries[0].contentRect;
      setSelectorWidth(width);
    });

    resizeObserver.observe(selectorElement);

    return () => {
      resizeObserver.unobserve(selectorElement);
      resizeObserver.disconnect();
    };
  }, []);
```
I think we can handle the resize in one single useEffect, no need to define a `updateSelectorWidth` function and no need to define a resizeObserver outside.

web/src/beta/lib/reearth-ui/components/Selector/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+  entries.forEach(entry => {
+    const selectorRef = entry.target;
+    if (selectorRef instanceof HTMLElement) {
+      const updateSelectorWidth = (selectorRef as any)._updateSelectorWidth;
+      if (updateSelectorWidth) {
+        updateSelectorWidth(selectorRef.clientWidth);
+      }
+    }
+  });
+});
```
airslice, line 29:
```suggestion
```
Not necessary i think, please check the comment on useEffect.

web/src/beta/lib/reearth-ui/components/Selector/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+  useEffect(() => {
+    setSelectedValue(value ?? (multiple ? [] : undefined));
+  }, [value, multiple]);
+
+  const updateSelectorWidth = useCallback(() => {
+    if (selectorRef.current) {
+      setSelectorWidth(selectorRef.current.clientWidth);
+    }
+  }, []);
```
airslice, line 57:
```suggestion
```
Not necessary i think, we can set the state value directly.

airslice (approved):

mkumbobeaty (approved):

m-abe-dev (approved):

airslice (approved):

PR #1005, opened 2024-06-04
ci: add `--force` option to push to protected branches

KeisukeYamashita:
# Overview

## What I've done

As the title says.

## How I tested

I need to merge this to test this out.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,21 +17,21 @@ jobs:
     runs-on: ubuntu-latest
     if: github.ref == 'refs/heads/release'
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -17,21 +17,21 @@ jobs:
     runs-on: ubuntu-latest
     if: github.ref == 'refs/heads/release'
     steps:
-      - name: Set up git config
```
KeisukeYamashita, line 20: > [!NOTE]
> I moved after checkout because we only need the configure these only if the checkout successes.

PR #1020, opened 2024-06-19
chore: remove deprecated husky command

mulengawilfred:
# Overview
`husky install ...` command is deprecated. Also, this is a small PR to make sure I can push commits hence the PR.


mulengawilfred:
Looks like `install` is still required so will close this one for now.

PR #1023, opened 2024-06-20
chore(web): visualizer dashbord

mkumbobeaty:
# Overview

- Implemented Visualizer Dashboard based on design
- Improved some libraries UI
- Added Loading UI in lib

## Memo
Screenshot 2024-06-25 at 10 08 22


web/src/beta/features/Assets/hooks.ts
```diff
@@ -95,7 +96,7 @@ export default ({
   }, [endCursor, sort, fetchMore, hasMoreAssets, isGettingMore]);
 
   const handleSortChange = useCallback(
-    (type?: string, reverse?: boolean) => {
+    (type?: string | string[], reverse?: boolean) => {
```
airslice, line 99:
```suggestion
    (type?: string, reverse?: boolean) => {
```
I think for `type` it should not be an array. If this comes from Selector, checking for string or string[] should be handled before calling this.

web/src/beta/features/Assets/hooks.ts
```diff
@@ -45,6 +45,7 @@ export default ({
   const [sort, setSort] = useState<{ type?: SortType; reverse?: boolean }>();
   const [searchTerm, setSearchTerm] = useState<string>();
   const [selectedAssets, selectAsset] = useState<Asset[]>([]);
+  const [selectedAssetId, setSelectedAssetId] = useState<string | undefined>(undefined);
```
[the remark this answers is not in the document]
airslice: Oh i find another `selectedAsset` in code, so is selectedAssets still necessary?

web/src/beta/features/Dashboard/ContentsContainer/Assets/AssetCard.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <Typography size="body">{asset.name}</Typography>
+      </AssetName>
+    </CardWrapper>
+  );
+};
+
+const CardWrapper = styled("div")<{ isSelected: boolean }>(({ theme, isSelected }) => ({
+  display: "flex",
+  flexDirection: "column",
+  border: `2px solid ${isSelected ? theme.select.main : "transparent"}`,
```
airslice, line 48: Why do we have a 2px border? I didn't find it on design.

web/src/services/i18n/translations/en.yml
```diff
@@ -48,6 +48,29 @@ A To Z: A To Z
 Z To A: Z To A
 Size Small to Large: Size Small to Large
 Size Large to Small: Size Large to Small
+Upload File: Upload File
+Assets: Assets
+'Sort:': Sort:'
```
airslice, line 53:
```suggestion
'Sort:': 'Sort:'
```
Do we really need `:` included?

web/src/beta/features/Dashboard/ContentsContainer/Projects/index.tsx
```diff
@@ -0,0 +1,197 @@
...
+  projectName?: string;
+  isHovered?: boolean;
+  onStarClick?: (e: MouseEvent<Element>, projectId: string) => void;
+  onProjectOpen?: () => void;
+  onProjectUpdate?: (project: ProjectType, projectId: string) => void;
+  onProjectSelect?: () => void;
+  onBlur?: () => void;
+  onChange?: (value: string) => void;
+  onDoubleClick?: () => void;
+  onHover?: (v: boolean) => void;
```
airslice, line 33: Could you update the namings of these events? Consider it's prop of `Layout` current naming will be understood as onLayoutBlur onLayoutChange etc. which seems confused.

airslice: And seems this type doesn't been used in this component, maybe better define it in an independent type file?

web/src/beta/features/Dashboard/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+export const tabsItem: Omit<TabItems[], "active"> = [
+  { id: "project", text: "Project", icon: "grid" },
+  { id: "asset", text: "Assets", icon: "file" },
+  { id: "members", text: "Members", icon: "appearance" },
+  { id: "bin", text: "Recycle bin", icon: "trash" },
+  { id: "plugin", text: "Plugin Playground", icon: "puzzlePiece", path: " " },
+  { id: "documentary", text: "Documentary", icon: "book", path: " " },
+  { id: "community", text: "Community", icon: "usersFour", path: " " },
+  { id: "help", text: "Help & Support", icon: "question", path: " " },
+];
```
airslice, line 25: I think it's better to define them seprately as topTabs and bottomTabs here, i saw you're doing somthing like .slice(3) .slice(4) with this list just to separete them into two parts, this way could easlisy lead to mistake when we change this list in the futrue.
And for the bottom, actually they are only links, maybe we should not regard them as tab.

web/src/beta/features/Dashboard/LeftSidePanel/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+              path={tab.path}
+              text={t(tab.text || "")}
+              icon={tab.icon}
+              active={tab.id === currentTab}
+            />
+          ))}
+        </>
+        <Version>
+          <Typography size="body" color={theme.content.weak}>
+            {t("Version 1.0 LTS")}
```
airslice, line 72: We need to get the version number from package.json. It's difficult for program to know whether it is LTS or not so i think we should not said `LTS` here, just using the actuall version number is enough.

web/src/beta/features/Dashboard/ContentsContainer/Projects/hooks.ts
```diff
@@ -0,0 +1,143 @@
...
+      ),
+    [],
+  );
+
+  const handleVisibility = useCallback(() => {
+    setVisible(!visible);
+  }, [visible]);
+
+  const handleViewChange = (newView: "grid" | "list") => {
+    localStorage.setItem("viewState", newView);
```
airslice, line 104: I think it's better to define a constant and export it for this storage key, using the same one when get/set it. The naming can be more specific for example `reearth-visualizer-dashboard-projects-view-state`

web/src/beta/features/Dashboard/ContentsContainer/Assets/AssetCard.tsx
airslice: Sorry i didn't mean it should not have a border, the key point is `2px`, i think in our current design usually is `1px` border.

airslice (approved):

PR #1026, opened 2024-06-20
chore: rename beta plugin api properties

mulengawilfred:
# Overview
This PR renames properties related to beta plugin api. @airslice , let's discuss about the changes when you have the time.
## What I've done
renamed properties in this file src/beta/features/Visualizer/Crust/Plugins/plugin_types.ts. Also added types where any type was used


mulengawilfred:
Current implementation did not match task requirements

PR #1036, opened 2024-06-26
feat(web): basic structure of plugin playground

airslice:
# Overview

This PR adds the basic structure of plugin playground to help further development.
Added new router as `/plugin-playground`.

Also refactors some on Tab and Area component.
- Tab
  - The switch tab logic should be handled internally and support controlled by props as well.
- Area
  - Updates some on vertical collapse.  

## Screenshot
localhost_3000_plugin-playground(Desktop 1920)

### DIFF
--- a/web/src/beta/features/PluginPlayground/Viewer/hooks.ts
+++ b/web/src/beta/features/PluginPlayground/Viewer/hooks.ts
@@ -1,0 +1,44 @@
+import { useEffect, useMemo, useRef, useState } from "react";
+
+import { Camera } from "@reearth/beta/utils/value";
+import { MapRef } from "@reearth/core";
+import { config } from "@reearth/services/config";
+
+export default () => {
+  // Refrence: hooks of Editor/Visualizer/hooks.ts and Publish/hooks.ts
+  const visualizerRef = useRef<MapRef | null>(null);
+  const [ready, setReady] = useState(false);
+  const [currentCamera, setCurrentCamera] = useState<Camera | undefined>(undefined);
+
+  const engineMeta = useMemo(
+    () => ({
+      cesiumIonAccessToken: config()?.cesiumIonAccessToken,
+    }),
+    [],
+  );
+
+  const sceneProperty = useMemo(
+    () => ({
+      tiles: [
+        {
+          id: "default",
+          type: "default",
+        },
+      ],
+    }),
+    [],
+  );
+
+  useEffect(() => {
+    setReady(true);
+  }, []);
+
+  return {
+    visualizerRef,
+    sceneProperty,
+    ready,
+    engineMeta,
+    currentCamera,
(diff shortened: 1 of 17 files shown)


web/src/beta/ui/layout/Area/index.tsx
```diff
@@ -235,22 +235,19 @@ export const Area = forwardRef<AreaRef, AreaProps>(
         localStorage.setItem(collapsedStorageKey, "0");
       }
       setSize(prevSize => {
-        const width = direction === "column" ? DEFAULT_WIDTH : prevSize.width;
-        const height = direction === "row" ? DEFAULT_HEIGHT : prevSize.height;
+        const w = direction === "column" ? width ?? DEFAULT_WIDTH : prevSize.width;
+        const h = direction === "row" ? height ?? DEFAULT_HEIGHT : prevSize.height;
```
mulengawilfred, line 239: Maybe we can use something more descriptive like `previousWidth` or `prevWidth` and `previousHeight` or `prevHeight` in this case?

web/src/beta/features/PluginPlayground/Viewer/hooks.ts
```diff
@@ -0,0 +1,44 @@
...
+          type: "default",
+        },
+      ],
+    }),
+    [],
+  );
+
+  useEffect(() => {
+    setReady(true);
+  }, []);
```
mulengawilfred, line 34: Sorry, I am trying to understand why we `setReady` to `true` here?

mulengawilfred (approved):
Marked as 'approved' as there are minor comments made which don't affect functionality.

web/src/beta/ui/layout/Area/index.tsx
airslice: This `width` or `height` here is from component props, which means component user gives an initialWidth or initialHeight, so using `initialWidth` `initialHeight` might be better. I'll update this. Thank you!

mulengawilfred: Got it.

web/src/beta/features/PluginPlayground/Viewer/hooks.ts
airslice: `ready` is passing to core Visualizer and has a long history i guess. To my understanding it can be used to make sure data is ready before engine really renders. I left this logic here just in case in the future you need to make sure some data is there, but for now i think remove it and directly passing a ready=true also works.

mkumbobeaty (approved):

web/src/beta/features/PluginPlayground/Viewer/hooks.ts
airslice: I'll mark this as resolved since this part is just a temp implementation.

PR #1049, opened 2024-07-03
fix(web): embed git commit hash from vite in correct way

keiya01:
# Overview

I added a functionality to embed git commit hash from vite define API in [this commit](https://github.com/reearth/reearth-visualizer/pull/1047), but the variable is removed in production build, so I fixed to assign the variable into `window` in runtime.

### DIFF
--- a/web/src/services/config/index.ts
+++ b/web/src/services/config/index.ts
@@ -51,10 +51,12 @@ export type Config = {
 
 declare global {
   let __APP_VERSION__: string;
+  let __REEARTH_COMMIT_HASH__: string;
   interface Window {
     REEARTH_CONFIG?: Config;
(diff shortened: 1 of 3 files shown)


web/src/services/config/index.ts
```diff
@@ -51,10 +51,12 @@ export type Config = {
 
 declare global {
   let __APP_VERSION__: string;
```
rot1024, line 53: not used? hash = version?

rot1024 (approved):

web/src/services/config/index.ts
keiya01: It's package version as application, so it's not related to the hash.

PR #1051, opened 2024-07-09
chore: migrate account and workspace pages to beta folder

mulengawilfred:
# Overview
As per title. Also moved a few components and refactored a few that gave errors.


web/src/services/theme/publishedAppProvider.tsx
```diff
@@ -1,15 +1,14 @@
...
-import betaTheme from "./reearthTheme/darkTheme";
+import darkTheme from "./reearthTheme/darkTheme";
+import lightTheme from "./reearthTheme/lightTheme";
 
 const Provider: React.FC<{ children?: ReactNode }> = ({ children }) => {
   const theme = {
-    classic: classicTheme,
-    ...betaTheme,
+    ...lightTheme,
+    ...darkTheme,
```
airslice, line 11:
```suggestion
    ...darkTheme,
```
Seems we only use darkTheme here (i don't know the reason though). Let's keep the behavior as-is for this PR.

mulengawilfred: Yes, the emotion theme provider gave an type error if I tried to specify properties to it like in the screenshot. I think we will need to refactor this when we decide how to add styles. I think TailwindCSS would be a good one for us.
Screenshot 2024-07-17 at 12 15 52 PM

airslice: For published page the theme here might not be in actual use. Just give it darkTheme as we did before is okey i think. What you did on screen shot changed the type and it will lead to type error for sure.
Tailwind & styled component both have their own pros and cons, both popular anyway. Maybe we will change our choice when necessary but that's another topic.

mulengawilfred: Got it. I misunderstood. I've removed the light theme and kept dark theme.

mkumbobeaty (approved):

airslice (approved):

PR #1058, opened 2024-07-17
refactor(web): scene settings inspector

mkumbobeaty:
# Overview
-Refactor scene settings inspector to use current beta UI fields
## What I've done

- refactor property fields to use current fields component
- use Number input field for all slider field while waiting for slider design to be complete
- refactor scene settings inspector panel

## What I haven't done
- slider fields and range field component not yet have clear design so not using them in property field


web/src/beta/ui/fields/Properties/ProprtyField.tsx
```diff
@@ -0,0 +1,156 @@
...
+          description={schema.description}
+          onChange={handleChange}
+        />
+      ) : schema.type === "latlng" ? (
+        <TwinInputField
+          key={schema.id}
+          commonTitle={schema.name}
+          values={[(value as LatLng)?.lat, (value as LatLng)?.lng]}
+          description={schema.description}
+          onChange={handleChange}
```
airslice, line 126: I think it's better to have onBlur as well.

web/src/beta/ui/fields/Properties/ProprtyField.tsx
```diff
@@ -0,0 +1,156 @@
...
+        />
+      ) : schema.type === "spacing" ? (
+        <SpacingField
+          key={schema.id}
+          commonTitle={schema.name}
+          value={(value as SpacingValues) ?? ""}
+          description={schema.description}
+          min={schema.min}
+          max={schema.max}
+          onChange={handleChange}
```
airslice, line 99: I think it's better to have onBlur as well.

airslice (approved):

PR #1070, opened 2024-07-25
chore(web): upgrade vite to 5.3.4

mulengawilfred:
# Overview
Upgrades project version of vite to 5.3.4
## What I've done
Ran `yarn add --dev vite`  command to upgrade and tested the frontend to see if any issues occurred and none were found. However, looks like we are not catching a lot of typescript errors with our current TS config which I can take up in a separate PR.


mulengawilfred:
Closing this PR as a fix is already in place.

PR #1078, opened 2024-07-29
chore(web): update web README.md and local env

mulengawilfred:
# Overview
Not a complete update. If there is anything that can be included please add a comment. Would be good to add as much information so that new members ( sometimes even old members ) can know how to run the project and anything else they would need to know. 
## Summary by CodeRabbit

- **New Features**
	- Introduced new environment configuration variables to enhance the application's setup, including developer mode, API endpoints, and authentication settings.

- **Chores**
	- Updated the `.gitignore` to specifically ignore the `.env` file, allowing for more precise management of environment configurations.
	- Removed the `developerMode` property from the configuration, streamlining settings for production use.


web/.env.local
```diff
@@ -0,0 +1,7 @@
+REEARTH_WEB_DEVELOPER_MODE=
```
airslice, line 1:
```suggestion
```
I think this one is only for show create beta project button on dashboard, which is not exists any more, could you help double confirm whether it is still useful?

web/.env.local
```diff
@@ -0,0 +1,7 @@
...
+REEARTH_WEB_CESIUM_ION_TOKEN_URL=
```
airslice, line 2:
```suggestion
```
This one is more for us only, it's not needed for OSS i think.

web/.env.local
```diff
@@ -0,0 +1,7 @@
...
+REEARTH_WEB_API=
+REEARTH_WEB_PLUGINS=
+REEARTH_WEB_AUTH0_DOMAIN=
+REEARTH_WEB_AUTH0_AUDIENCE=
+REEARTH_WEB_AUTH0_CLIENT_ID=
```
airslice, line 7: Maybe we can provide the local urls as well as the template?

web/.env.local
mulengawilfred: Looks like this one is used in `e2e/utils/config.ts`

web/.env.local
mulengawilfred: Yes we can. Please share them or push directly to the PR.

web/.env.local
mulengawilfred: Okay, I have indicated which ones are required and which ones are optional

web/.env.local
airslice: I think it's been configed but not in use any more, i removed it.

web/.env.local
airslice: I add the url for first two, the rest can be added after we fix the local auth system i think

PR #1199, opened 2024-10-25
fix(web): improve project import 

mkumbobeaty:
# Overview
- Add teamId in project import  mutation input
## Summary by CodeRabbit

- **New Features**
	- Enhanced project import functionality by allowing imports with both `teamId` and `file`.
- **Bug Fixes**
	- Simplified error handling for project export operations, improving user feedback.
- **Documentation**
	- Updated GraphQL mutation signatures to reflect new input requirements for project imports.


web/src/beta/features/Dashboard/ContentsContainer/Projects/hooks.ts
```diff
@@ -239,13 +239,16 @@ export default (workspaceId?: string) => {
     async (event: React.ChangeEvent<HTMLInputElement>) => {
       const file = event.target.files?.[0];
       if (file) {
-        const result = await useImportProject(file);
+        const result = await useImportProject({
+          teamId: workspaceId || "",
```
mulengawilfred, line 243: Sorry would need some help clarifying the reason we chose to use `workspaceId` as part of the input params. Since this function takes in the file from the user's local device as an input, I am wondering how the `workspaceId` will be used to improve the input. Is it so that we know which workspace the import is under?

web/src/beta/features/Dashboard/ContentsContainer/Projects/Project/hooks.ts
```diff
@@ -70,13 +70,7 @@ export default ({
...
 
-    const result = await useExportProject(project.id);
-
-    if (result.status === "success") {
-      console.log("export success");
-    } else {
-      console.error("Failed to export project:", result.status);
-    }
+    await useExportProject(project.id);
   }, [useExportProject, project.id]);
```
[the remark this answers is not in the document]
mulengawilfred: Sorry, I am wondering why we would want to remove the error handling here.

mulengawilfred:
Implementation is good. I have two questions added and I think we would also require backend changes in this repo to support the new input property.

web/src/beta/features/Dashboard/ContentsContainer/Projects/hooks.ts
mkumbobeaty: @mulengawilfred  the API requires to input which are teamId and file and the teamId is same as workspaceId you can check the importProject mutation

mkumbobeaty:
> Implementation is good. I have two questions added and I think we would also require backend changes in this repo to support the new input property.

mkumbobeaty:
> Implementation is good. I have two questions added and I think we would also require backend changes in this repo to support the new input property.

@mulengawilfred actually the backend already handled this input that why I updated on front-end

web/src/beta/features/Dashboard/ContentsContainer/Projects/hooks.ts
airslice: My question is: is "" valid here? Or should we prevent the import if we don't get the workspaceId?

mulengawilfred: Ah okay. I didn't see that. Thanks @mkumbobeaty  🙏

airslice (approved):

PR #1251, opened 2024-11-14
fix: camera returning to initial camera position after scene settings change

mulengawilfred:
# Overview
Short description on how to reproduce the bug: 
- When the initial camera position is set under the `Scene` -> `Camera` settings, and then the camera position is changed by dragging the globe. Then if a change is made to a scene setting e.g enabling globe lighting, the camera automatically goes back to the initial camera position. 

- I removed the useEffect that updates the currentCamera value with the initialCamera value. 
## How I tested
Tested the steps in the Overview after and before the change made
## Summary by CodeRabbit

- **New Features**
	- Enhanced camera state management within the visualizer component, ensuring updates occur only after initialization for improved user experience.
  
- **Bug Fixes**
	- Streamlined camera state updates to prevent unnecessary changes during initial renders, allowing for better user control over camera settings.


web/src/beta/features/Editor/Visualizer/hooks.ts
mulengawilfred, line 1: on line 111 we update the initialCamera value when a scene property change which if together with the useEffect on line 125, the currentCamera value will also be updated. Looks like the currentCamera value gets updated by `handleCameraUpdate` on 80 and is passed to Visualizer so we might not require the useEffect but could have a use-case. Would need to double-confirm.

airslice: I think what you deleted is useful, but the set action should be trigged only once, i think we can keep that and add a inited to avoid unnecessary trigger.
```
  const inited = useRef(false);

  useEffect(() => {
    if (inited.current) return;
    inited.current = true;
    setCurrentCamera(initialCamera);
  }, [initialCamera, setCurrentCamera]);
  ```

airslice (approved):

PR #1257, opened 2024-11-15
chore(web): add type check in pre-commit hook

mulengawilfred:
# Overview
Updated the pre-commit hook to run  `yarn run type` before the `lint-staged` command so that text that needs to be translated and type check can be done before commiting. This will hopefully reduce chances of it being caught in the ci-build step. 
## What I've done
Tested locally by creating a file with eslint and type errors and made changes to text copy to see if the command works.
Screenshot 2024-11-15 at 10 15 34 AM
## How I tested
As mentioned above.
## Summary by CodeRabbit

- **Chores**
	- Updated the pre-commit hook to include type checking before linting, enhancing the code quality checks prior to commits.

### DIFF
--- a/web/.husky/pre-commit
+++ b/web/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-cd web && yarn run lint-staged && yarn run type && yarn run i18n
+cd web && yarn run i18n && yarn run type && yarn run lint-staged 
--- a/web/.husky/pre-commit
+++ b/web/.husky/pre-commit
@@ -1,4 +1,5 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-cd web yarn run type && yarn run lint-staged 
+yarn --cwd web run type
+yarn --cwd web run lint-staged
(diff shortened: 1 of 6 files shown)


web/.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn --cwd web run lint-staged
+cd web yarn run type && yarn run lint-staged
```
airslice, line 4: Is there any special reason for changing `yarn --cwd` to `cd web yarn`? I think it can work in both way, but yarn --cwd is more self-contained, it doesn't alter shell's working directory and could avoid side effects of that. (although we only have one cmd in this case but i think it's still ideal)

mulengawilfred: Thanks. You're right, `yarn --cwd` doesn't change the working directory. I've updated to use that instead. Please note that the command has been repeated as it might not be possible to chain the commands into one ( need to specify `yarn --cwd` in the second command as well ).

airslice (approved):

web/.husky/pre-commit
airslice: Got it, thank you!

PR #1260, opened 2024-11-19
fix(web): terrain not applied properly on editor

airslice:
# Overview

When terrain is enabled it cannot be applied on map properly.
We didn't find the root cause, so this is just a workaround, temporarily disable it and enable it again in an async function.

## Summary by CodeRabbit

- **New Features**
	- Introduced a temporary terrain rendering disable feature to improve loading performance in the editor.
	- Enhanced camera state management for better user experience.

- **Bug Fixes**
	- Resolved issues related to terrain not loading properly in the editor.


mulengawilfred (approved):
Thanks for finding the fix for this! Much appreciated!

mkumbobeaty (approved):

PR #1262, opened 2024-11-19
fix(web): 3d tiles' style cannot be reset

airslice:
# Overview

After you assign a style to 3dtiles and reset, the style can't be reset properly on map. That's due to the '3dtiles' property is `undefined`. We should give a default empty object `{}` as the default 3dtiles style.

## Summary by CodeRabbit

- **New Features**
	- Enhanced styling capabilities for 3D tiles, including new properties for dynamic color assignment and rendering options.
	- Updated visibility settings for marker and polygon styles.
- **Improvements**
	- Added support for additional properties in style definitions, improving customization options.


mulengawilfred (approved):
Thanks for the fix for this! Works as expected

mkumbobeaty (approved):

PR #1271, opened 2024-11-21
chore(web): add prettier package to organize imports

mulengawilfred:
# Overview
Small package that will help organize imports when a file is saved in the editor. It does not scan through all the files but only checks when a single file is open. Might not be needed but helps keep imports in order and one less thing for the developer to think of how to sort. 
## Summary by CodeRabbit

- **New Features**
	- Introduced a new Prettier plugin to automatically organize imports in the codebase.
- **Chores**
	- Updated the project version to `1.0.0-alpha.8`.
	- Added `prettier-plugin-organize-imports` as a new development dependency.

### DIFF
--- a/web/.prettierrc
+++ b/web/.prettierrc
@@ -1,3 +1,4 @@
 {
-  "trailingComma": "none"
+  "trailingComma": "none",
+  "plugins": ["prettier-plugin-organize-imports"]
 }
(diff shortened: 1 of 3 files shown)


mulengawilfred:
Coderrabbitai says I made a change to the version to `1.0.0-alpha.8` but I dont think I have

mulengawilfred:
Using ESLint import/order

PR #1282, opened 2024-11-26
fix(web): plugin playground/ Modify the behavior of the HTML Editor modal

devshun:
# Overview

modify the behavior of the HTML Editor modal

## What I've done
Before:
Activate the button to open the modal if there is a valid string (e.g., reearth.ui.show, reearth.popup.show, etc.) in the code.

After:
Activate the button to open the modal only if there is a valid string (e.g., reearth.ui.show, reearth.popup.show, etc.) currently under the cursor.

## Summary by CodeRabbit

- **New Features**
  - Enhanced HTML injection handling in the Code component.
  - Introduced a new optional `onMount` prop for the `CodeInput` component to allow actions upon editor initialization.

- **Bug Fixes**
  - Improved state management and rendering logic for the HTML editor modal.

- **Chores**
  - Removed unused functions related to HTML extraction and injection to streamline code.


web/src/beta/features/PluginPlayground/Code/index.tsx
```diff
@@ -16,26 +13,88 @@ type Props = {
...
+    (value: string, offset: number): RegExpExecArray | null => {
+      for (const pattern of REEARTH_HTML_INJECTION_PATTERNS) {
+        const match = pattern.exec(value);
+        if (match) {
+          const start = match.index;
+          const end = start + match[0].length;
+          if (offset > start && offset < end) {
+            return match;
+          }
+        }
```
airslice, line 46:
```suggestion
        const matches = value.matchAll(pattern);
        for (const match of matches) {
          if (match) {
            const start = match.index;
            const end = start + match[0].length;
            if (offset > start && offset < end) {
              return match;
            }
          }
        }
```
Multiple call of the same API is allowed, i guess we need to support it as well, for example there could be multiple reearth.modal.show written in the code.

mkumbobeaty (approved):

web/src/beta/features/PluginPlayground/Code/index.tsx
devshun: thanks!
I fixed this commit 

https://github.com/reearth/reearth-visualizer/pull/1282/commits/ff14cbe02d091cf7dface4fd4e4aa3d22fdc2a55

airslice (approved):

PR #1326, opened 2024-12-24
ci(server): mongo update 7

hexaforce:
# Overview

### I will update the version of MongoDB used for development to version 7. 

This will match the current cloud environment.

Additionally, I will set up a replica for MongoDB. This is necessary to enable complex transactions in MongoDB and is required when running reearth-cloud locally.

This will not affect reearth-visualizer.

## Summary by CodeRabbit

- **Chores**
	- Updated the MongoDB service definition in the CI workflow.
	- Added caching for dependencies in the CI configuration.


mongo-init.js
pyshx, line 1: Do we need this?

hexaforce: command: ["mongod", "--replSet", "rs0"] 

The idea is to set up the replica (`rs0`) using the "mongod" startup option, and once it starts, the `mongo-init.js `will be executed, enabling `rs0`.

pyshx: We don't need replica, we can keep on using mongo the way we've been doing so far.

hexaforce: Got it. I will remove it.
When running reearth-cloud locally, it needs to be configured individually.

PR #1336, opened 2025-01-07
chore(server): include Makefile.local on Makefile

soneda-yuya:
# Overview
I think I want to use go hot reload feature.
But I concerned it is not suitable for oss to depend on specified hot reload feature. 
So I have idea which developer create Makefile for local and include it on main Makefile.

I want to add the following description to Makefile.local to realize the hotreload function locally

```
# behavior not confirmed in windows.
# reflex is hot reloading tool for go.
REFLEX_BIN := $(shell which reflex)
install-reflex:
ifndef REFLEX_BIN
	@echo "reflex is not installed. Installing..."
	@go install github.com/cespare/reflex@v0.3.1
else
	@echo "reflex is already installed."
endif

dev: install-reflex
	reflex -r '\\.go$$' -s go run ./cmd/reearth
```

## What I've done
-  include Makefile.local on Makefile
- add Makefile.local on `.gitignore`

- Whether this way is appropriate

## Summary by CodeRabbit

- **Chores**
	- Updated `.gitignore` to exclude `Makefile.local`
	- Modified Makefile to optionally include a local configuration file


soneda-yuya:
I close this PR, because this problem resolve this [PR](https://github.com/reearth/reearth-visualizer/pull/1337).

PR #1345, opened 2025-01-10
chore(server): enable to set visualizer db name by .env

soneda-yuya:
# Overview
There are problem we can't set database of mongo db by env file.

## What I've done
- if you set value on REEARTH_DB_VISUALIZER variable of .env,  you can switch database defined on .env value.

### DIFF
--- a/server/internal/app/config/config.go
+++ b/server/internal/app/config/config.go
@@ -30,6 +30,7 @@ type Config struct {
 	Dev              bool              `pp:",omitempty"`
 	DB               string            `default:"mongodb://localhost"`
 	DB_Account       string            `pp:",omitempty"`
+	DB_Visualizer    string            `pp:",omitempty"`
 	DB_Users         []appx.NamedURI   `pp:",omitempty"`
 	GraphQL          GraphQLConfig     `pp:",omitempty"`
 	Published        PublishedConfig   `pp:",omitempty"`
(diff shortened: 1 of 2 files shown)


server/.env.example
```diff
@@ -1,6 +1,8 @@
 # General
 PORT=8080
 REEARTH_DB=mongodb://localhost
+REEARTH_DB_ACCOUNT=reearth # for account database
```
pyshx, line 4: we don't need these comments.

server/internal/app/config/config.go
```diff
@@ -30,6 +30,7 @@ type Config struct {
 	Dev              bool              `pp:",omitempty"`
 	DB               string            `default:"mongodb://localhost"`
 	DB_Account       string            `pp:",omitempty"`
+	DB_Vis           string            `pp:",omitempty"`
```
pyshx, line 33: Please place this alphabetically.

server/internal/app/repo.go
```diff
@@ -42,11 +42,11 @@ func initReposAndGateways(ctx context.Context, conf *config.Config, debug bool)
 		log.Fatalf("mongo error: %+v\n", err)
 	}
 
-	// repos
+	// for account database
 	accountDatabase := conf.DB_Account
 	accountRepoCompat := false
 	if accountDatabase == "" {
-		accountDatabase = databaseName
+		accountDatabase = defaultDatabase
```
pyshx, line 49: Please add a `deafultAccountDatabase = "reearth-account"`

pyshx (approved):

pyshx (approved):

PR #1356, opened 2025-01-16
chore(web): support create initial camera with Edit panel

mkumbobeaty:
# Overview
In the previous implementation, we were unable to manually create the initial camera setup using the Edit Panel. In this PR, I improved some of the logic to allow setting up the initial camera directly in the Edit Panel.
## What I've done
- support create initial camera with Edit panel
- I set the default FOV to 1, as specified in the yml file, since it is expected to be between 0.1 and 2.0.

Screenshot 2025-01-16 at 11 42 32

## Summary by CodeRabbit

- **New Features**
  - Introduced default camera configuration to ensure consistent camera property handling.
  
- **Bug Fixes**
  - Resolved potential undefined camera state by implementing default camera values.
  - Improved input field behavior when camera properties are not explicitly set.
  - Enhanced timing for camera field of view updates to allow smoother transitions.


mulengawilfred (approved):
Tested on main and on this PR. Looks like the initial camera position can now be set. Approved!

airslice (approved):

PR #1365, opened 2025-01-21
feat(web): add hide flyto and delete layer plugin example 

mkumbobeaty:
# Overview
- Added example plugin for hide, flyto and delete layer in layer management
## What I've done

Screenshot 2025-01-21 at 18 10 14

## Summary by CodeRabbit

- **New Features**
	- Added a new plugin for managing layers with enhanced interaction capabilities.
	- Introduced functionality to hide, fly to, and delete layers dynamically.
	- Implemented layer visibility controls and management options.

- **Improvements**
	- Updated the `presetPlugins` to include the new `hideFlyToDeleteLayer` plugin.
	- Enhanced layer interaction with more intuitive user interface elements.


web/src/beta/features/PluginPlayground/Plugins/presets/layers/hideFlyToDeleteLayer.ts
```diff
@@ -0,0 +1,240 @@
...
+  }
+};
+
+reearth.layers.add(layerGeojson);
+reearth.layers.add(layer3dTiles);
+
+const layers = reearth.layers.layers
+
+// filter layers
+const presetLayerIds = ["1", "2"];
```
mulengawilfred, line 69: I think we could omit the preset layers as ability to hide and zoom to the layer is possible from the layers panel but one to confirm with @airslice

web/src/beta/features/PluginPlayground/Plugins/presets/layers/hideFlyToDeleteLayer.ts
```diff
@@ -0,0 +1,240 @@
...
+const presetLayers = layers.filter(layer => presetLayerIds.includes(layer.id));
+const pluginLayers = layers.filter(layer => !presetLayerIds.includes(layer.id));
+
+const generateLayerItem = (layer, isPreset) => {
+  return \`
+    <li>
+      <span id="layer-name">\${layer.title}</span>
+      <div class="actions">
+        <input 
+          type="checkbox"
```
mulengawilfred, line 79: could use the `crosshair` icon here to symbolize that it clicking it has a zoom functionality.

web/src/beta/features/PluginPlayground/Plugins/presets/layers/hideFlyToDeleteLayer.ts
airslice: Yes we could do hide/zoom to the preset layer on layers panel, but here for this plugin what we want to do is show our user what we can do / we will get from plugin API. Preset layers in plugin playground is similar to the layers user added to a project from Editor regarding a real reearth project, therefore when plugin api layers.layers been called it also could get both layers from preset (project) and from plugin added. It's more clear for user to understand i think.

mulengawilfred (commented):
Added a few comments

web/src/beta/features/PluginPlayground/Plugins/presets/layers/hideFlyToDeleteLayer.ts
```diff
@@ -0,0 +1,240 @@
...
+    type: "geojson",
+    url: "https://reearth.github.io/visualizer-plugin-sample-data/public/geojson/sample_polygon_polyline_marker.geojson"
+  },
+  polygon: {
+    fillColor: "red"
+  },
+  polyline: {
+    strokeColor: "red"
+  },
+  marker: {
```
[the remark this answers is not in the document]
airslice: @mulengawilfred i think we need to provide the `marker` property even value as an empty object if we want to render markers for this layer.

web/src/beta/features/PluginPlayground/Plugins/presets/layers/hideFlyToDeleteLayer.ts
airslice: Ah sorry it was an icon, we want to shorten the code (we don't need embed svg in this case) and make it simple so changed to text.

web/src/beta/features/PluginPlayground/Plugins/presets/layers/hideFlyToDeleteLayer.ts
```diff
@@ -0,0 +1,240 @@
...
+  },
+  "3dtiles": {
+    // Settings for the 3D Tiles style.
+    pbr: false, 
+    selectedFeatureColor: "red" // If you select a feature, it will change color
+  }
+};
+
+reearth.layers.add(layerGeojson);
+reearth.layers.add(layer3dTiles);
```
airslice, line 64:
```suggestion
const pluginGeojsonLayerId = reearth.layers.add(layerGeojson);
const plugin3dTilesLayerId = reearth.layers.add(layer3dTiles);
```
We should be able to get the plugin layer ids in this way, and when do filter we don't need to hardcode the `1` `2` for preset layer ids.

airslice (approved):

PR #1420, opened 2025-02-12
refactor(web): support import plugin with zip file instead of single js file in plugin playground

mulengawilfred:
# Overview
Refactors logic to support zip file upload for plugins
## What I've done
- Added logic to restrict file upload to only a single zip file.
- Added logic to read the uploaded zip file and check if it contains only `.yml` and `.js` files and show and error if it does not.
- Re-write files under the `custom` folder with the files from the zip file.
- Restrict file upload to `custom` plugin as preset plugins should not be overriden.
## Summary by CodeRabbit

- **New Features**
	- Updated the plugin file upload process: users now upload a single zip archive containing valid plugin files. The system verifies the archive’s content and notifies users if the zip is empty or includes unsupported files before adding the plugins.
- **Improvements**
	- Updated titles for clarity in the zoom and layer management plugins.
- **Chores**
	- Removed an unnecessary import related to layer styling examples.


web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
```diff
@@ -170,42 +170,56 @@ export default () => {
...
+          const newPlugin = {
+            id: uuidv4(),
+            title: file.name,
+            files: pluginFiles
+          };
+
+          setPlugins((plugins) => [...plugins, newPlugin]);
+          setSelectedPluginId(newPlugin.id);
+          setSelectedFileId(newPlugin.files[0].id);
+        });
```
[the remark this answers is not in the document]
mulengawilfred: overwriting is intended

web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
```diff
@@ -170,42 +170,56 @@
...
+      zip.loadAsync(file).then((zip) => {
+        const files = Object.values(zip.files).filter((file) => !file.dir);
+        const filePromises = files.map((file) => file.async("text"));
+
+        Promise.all(filePromises).then((fileContents) => {
+          const pluginFiles = fileContents.map((content, index) => ({
+            id: uuidv4(),
+            title: files[index].name.split("/")[1],
+            sourceCode: content
+          }));
```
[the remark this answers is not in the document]
mulengawilfred: This should be okay for now

web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
```diff
@@ -170,42 +170,56 @@
...
-        )
-      );
-    };
-    reader.readAsText(file);
-  });
+      const zip = new JSZip();
+      zip.loadAsync(file).then((zip) => {
+        const files = Object.values(zip.files).filter((file) => !file.dir);
+        const filePromises = files.map((file) => file.async("text"));
+
```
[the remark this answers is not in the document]
mulengawilfred: Thanks. This has been added

web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
```diff
@@ -170,42 +170,64 @@ export default () => {
...
+                file.title.endsWith(".yml") || file.title.endsWith(".js")
+            );
+
+            if (!hasValidFiles) {
+              setNotification({
+                type: "error",
+                text: "Zip file must only contain .yml and .js files"
+              });
+              return;
+            }
```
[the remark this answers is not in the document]
mulengawilfred: Thanks.That makes sense.
@airslice , let me know if we should check the zip file size and what size is sounds appropriate as well as what number of files in the zip file a user should not exceed. Thank you!

web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
```diff
@@ -170,42 +170,71 @@ export default () => {
...
+            const newPlugin = {
+              id: uuidv4(),
+              title: file.name,
+              files: pluginFiles
+            };
+
+            setPlugins((plugins) => [...plugins, newPlugin]);
+            setSelectedPluginId(newPlugin.id);
+            setSelectedFileId(newPlugin.files[0].id);
+          })
```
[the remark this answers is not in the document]
mulengawilfred: Will keep it as is.

web/src/beta/features/PluginPlayground/Plugins/index.tsx
```diff
@@ -67,11 +67,13 @@ const Plugins: FC<Props> = ({
...
-          icon="import"
-          onClick={handleFileUpload}
-        />
+        {selectedPlugin.id === "my-plugin" && (
+          <IconButton
+            appearance="simple"
+            icon="import"
+            onClick={handleFileUpload}
+          />
+        )}
```
airslice, line 76: It might be a little bit difficult for user to notice that he can import plugin if this button is hidden when browse other plugins. I think we can keep the button avaliable always, and after import we automatically select MyPlugin.

web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
```diff
@@ -170,42 +170,77 @@ export default () => {
...
-    }
-    const result = validateFileTitle(
-      file.name,
-      selectedPlugin.files.map((f) => f.title)
-    );
-    if (!result.success) {
-      setNotification({ type: "error", text: result.message });
-      return;
-    }
+  const handleFileUpload = useFileInput(
```
airslice, line 173:
```suggestion
  const handlePluginImport = useFileInput(
```
Suggest to use a more clear naming here.

web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
```diff
@@ -170,42 +170,77 @@ export default () => {
...
+    (fileList) => {
+      const file = fileList?.[0];
+      if (!file) {
+        setNotification({ type: "error", text: "File not found" });
+        return;
+      }
 
-    const reader = new FileReader();
+      if (file.type !== "application/zip") {
+        setNotification({ type: "error", text: "Invalid file type" });
```
airslice, line 182: Maybe better have some description on error msg let user know it expects a single zip file.

web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
```diff
@@ -170,42 +170,77 @@ export default () => {
...
+        if (files.length === 0) {
+          setNotification({ type: "error", text: "Zip file is empty" });
+          return;
+        }
+
+        // Prevent path traversal
+        if (files.some((file) => file.name.includes(".."))) {
+          setNotification({ type: "error", text: "Invalid file paths in zip" });
+          return;
+        }
```
airslice, line 199: When will this happen? And what if user do have two points written in filename?

web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
```diff
@@ -170,42 +170,77 @@ export default () => {
...
+            const pluginFiles = fileContents.map((content, index) => ({
+              id: uuidv4(),
+              title: files[index].name.split("/")[1],
+              sourceCode: content
+            }));
+
+            const hasValidFiles = pluginFiles.every(
+              (file) =>
+                file.title.endsWith(".yml") || file.title.endsWith(".js")
+            );
```
airslice, line 214: Maybe we can skip this check. If the plugin misses any necessary file it should have an error message when extrude code.

web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
```diff
@@ -170,42 +170,77 @@ export default () => {
...
+            }
+
+            const newPlugin = {
+              id: uuidv4(),
+              title: file.name,
+              files: pluginFiles
+            };
+
+            setPlugins((plugins) => [...plugins, newPlugin]);
+            setSelectedPluginId(newPlugin.id);
```
airslice, line 231: This `setSelectedPlugin` can't work well since this new plugin is not exist on category, once you click `MyPlugin` it will revert to the previous MyPlugin.
Currently i'm think about making a new category `Imported`, as a dynamic category, similar to `Shared`, we can keep adding imported plugin into that category, and do select, what do you think?

web/src/beta/features/PluginPlayground/Plugins/index.tsx
mulengawilfred: Got it. I have reverted this change and if a user selects another preset plugin and tries to upload a plugin, the files under `custom` subfolder get overwritten and not the preset plugin.

web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
mulengawilfred: Thanks for spotting that. Done

web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
mulengawilfred: I think it was to prevent issues of creating a subfolder but looks like it doesnt create if there are two dots in the file name. I have removed this change.

web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
mulengawilfred: Thanks. I have updated the error message.

web/src/beta/features/PluginPlayground/Plugins/usePlugins.ts
airslice: Okey!

airslice (approved):

PR #1431, opened 2025-02-17
fix(server): attempt to fix golang lint file error

mulengawilfred:
# Overview
This PR attempts to fix the `ci-server-lint` error experienced in the CI build. From research looks like there is a change to the indentation required for `golangci.yml` regarding `goimports`. Will need backend team to confirm if this change is correct.

Screenshot of error 
Screenshot 2025-02-17 at 9 43 17 AM

### DIFF
--- a/server/.golangci.yml
+++ b/server/.golangci.yml
@@ -6,5 +6,6 @@ linters:
     - gofmt
     - goimports
 
-goimports:
-  local-prefixes: github.com/reearth/reearth-backend
+linters-settings:
+  goimports:
+    local-prefixes: github.com/reearth/reearth-backend


server/.golangci.yml
```diff
@@ -6,5 +6,6 @@ linters:
     - gofmt
     - goimports
 
-goimports:
-  local-prefixes: github.com/reearth/reearth-backend
+linters-settings:
+  goimports:
+    local-prefixes: github.com/reearth/reearth-backend
```
soneda-yuya, line 11: @mulengawilfred 
thanks for your improvement!
I believe this issue has existed for some time, but `github.com/reearth/reearth-backend` is from the Classic repository. Would it be possible to update it to `github.com/reearth/reearth/server` instead?

soneda-yuya (approved):
LGTM!

server/.golangci.yml
mulengawilfred: Thanks @soneda-yuya ! This change has been made. Thank you!

pyshx (approved):

PR #1471, opened 2025-03-06
feat(server): asset project association migration [VIZ-1381]

hexaforce:
# Overview
This migration associates the current assets with projects.

## What I've done

The target collections are as follows:

It performs a full scan of the database, retrieves assets based on their URL values, and assigns them to the corresponding projects.

"project"
"scene"
"nlsLayer"
"storytelling"
"style"
"property"
"propertySchema"
"plugin"

If the same asset is used by multiple sources, it will be assigned to the last one.

## Summary by CodeRabbit

- **New Features**
  - Enhanced asset management by introducing a new data migration that updates and synchronizes assets with their corresponding projects, ensuring improved data accuracy and consistency.
- **Tests**
  - Added automated tests to validate the asset-to-project linking process, ensuring reliable operation and reinforcing overall system stability.


pyshx (changes requested):
@hexaforce It should run if RUN_MIGRATION instead running through cmd now.

server/internal/infrastructure/mongo/migration/250305230545_asset_project_association.go
```diff
@@ -0,0 +1,266 @@
...
+
+	if err := cursor.Err(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func loadAssetURLPrefix() (string, error) {
+	if err := godotenv.Load(".env"); err != nil && !os.IsNotExist(err) {
```
pyshx, line 107: Can't it be obtained from config?

server/internal/infrastructure/mongo/migration/250305230545_asset_project_association.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	if err := cursor.Err(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func loadAssetURLPrefix() (string, error) {
```
pyshx, line 106: can't we pass the config directly from the migration app from main?

pyshx (approved):

PR #1528, opened 2025-03-27
fix(server): revert sort logic [VIZ-1423]

hexaforce:
# Overview

### The sort that was previously added for debugging was affecting the display order.
スクリーンショット 2025-03-27 10 46 32
スクリーンショット 2025-03-27 10 47 12

## Summary by CodeRabbit

- **New Features**
  - Introduced a new property identifier in scene details to enhance client data visibility.
  - Added a new test to validate the order of property fields in a scene's schema.

- **Refactor**
  - Improved the handling of property items and fields by aligning them with associated schemas.
  - Enhanced property operations to consistently integrate schema data, offering a more predictable presentation.
  - Simplified the output of property fields by removing sorting logic.

- **Bug Fixes**
  - Adjusted expected field order in tests to reflect changes in property field arrangements.

- **Chores**
  - Removed outdated test function to streamline the test suite.
  - Added a new function for end-to-end testing context.


server/pkg/property/schema_group.go
```diff
@@ -124,3 +118,29 @@ func (s *SchemaGroup) RepresentativeField() *SchemaField {
...
+	// Remove the field from the original position
+	s.fields = append(s.fields[:from], s.fields[from+1:]...)
+
+	// Insert the field into the new position
+	if to >= len(s.fields) {
+		s.fields = append(s.fields, field)
+	} else {
+		s.fields = append(s.fields[:to], append([]*SchemaField{field}, s.fields[to:]...)...)
+	}
+}
```
soneda-yuya, line 146: @hexaforce 
[nits]
Is this function only used for testing.
If so, would you mind moving it to the corresponding test file?

server/internal/adapter/gql/gqlmodel/convert_property.go
```diff
@@ -432,22 +441,42 @@ func ToPropertySchemaGroup(g *property.SchemaGroup, s id.PropertySchemaID) *Prop
...
+							fields = append(fields, gf)
+						}
+					}
+				}
+			}
+		}
+	} else {
+		fields = g.Fields(nil)
+	}
+
```
soneda-yuya, line 468: [imo]
The nesting is quite deep — would it be possible to split it into functions to reduce the nesting?
This is just a suggestion, so there’s no need to force it if it’s difficult to do.

server/pkg/property/schema_group.go
hexaforce: It's a feature to reorder properties.
Where would you like to move it?

server/internal/adapter/gql/gqlmodel/convert_property.go
hexaforce: That logic didn't work.
I performed a refactoring using ChatGPT.

soneda-yuya: Thanks.

server/pkg/property/schema_group.go
soneda-yuya: Oh I got it.
By writing it this way, I believe we can ensure the test content without overwriting the `manifest.json`.
Wouldn’t this allow us to remove the tests for `Move` and `E2ETestChange`?

```
package e2e

import (
	"log"
	"testing"

	"golang.org/x/text/language"
)

func TestPropertySchemaOrder(t *testing.T) {
	e := Server(t, baseSeeder)
	pId := createProject(e, "test")
	_, _, sId := createScene(e, pId)
	res := getScene(e, sId, language.Und.String())

	propID := res.Path("$.property.id").Raw().(string)
	groupID := res.Path("$.property.items[0].groups[0].id").Raw().(string)

	log.Println("propID: ", propID)
	log.Println("groupID: ", groupID)

	_, r := updatePropertyValue(e, propID, "tiles", groupID, "tile_type", "open_street_map", "STRING")
	r.Path("$.data.updatePropertyValue.propertyField.value").IsEqual("open_street_map")

	_, r = updatePropertyValue(e, propID, "tiles", groupID, "tile_opacity", 0.5, "NUMBER")
	r.Path("$.data.updatePropertyValue.propertyField.value").IsEqual(0.5)

	// 1. tile_type
	// 2. tile_opacity
	res = getScene(e, sId, language.Und.String())
	fields := res.Path("$.property.items[0].groups[0].fields").Array()
	fields.Value(0).Object().Value("fieldId").IsEqual("tile_type")
	fields.Value(1).Object().Value("fieldId").IsEqual("tile_opacity")

	// 1. tile_type
	// 2. tile_url
	// 3. tile_zoomLevel
	// 4. tile_opacity
	_, r = updatePropertyValue(e, propID, "tiles", groupID, "tile_url", "https://example.com", "STRING")
	r.Path("$.data.updatePropertyValue.propertyField.value").IsEqual("https://example.com")
	_, r = updatePropertyValue(e, propID, "tiles", groupID, "tile_zoomLevel", []int{10, 20}, "ARRAY")
	r.Path("$.data.updatePropertyValue.propertyField.value").IsEqual([]*int{nil, nil})

	res = getScene(e, sId, language.Und.String())
	fields = res.Path("$.property.items[0].groups[0].fields").Array()
	fields.Value(0).Object().Value("fieldId").IsEqual("tile_type")
	fields.Value(1).Object().Value("fieldId").IsEqual("tile_url")
	fields.Value(2).Object().Value("fieldId").IsEqual("tile_zoomLevel")
	fields.Value(3).Object().Value("fieldId").IsEqual("tile_opacity")
}

```

hexaforce: That makes the test meaningless.

I want to check if the order of the properties changes when the order is changed in the original manifest.yml.
As an example, I swapped tile_opacity and tile_zoomLevel to see if "tile_opacity" is returned before "tile_zoomLevel".

soneda-yuya: @hexaforce 
Yes, you are right.
On the other hand, we’d prefer to avoid including functions used only for testing in the domain layer if possible.

In Golang, everything outside the internal directory is basically exposed to external modules.
So, I felt a bit of discomfort having E2E-related concerns within that layer.
That’s why I was thinking—how about preparing a function like `OverridePluginManifest` instead?
By doing that, we can replace the manifest from the outside, and also separate the logic from concerns that belong in the internal package.

Also, since this involves manipulating a global variable, I’d like to add a lock if necessary.

By doing this, we should be able to write test code where needed and run tests without having to provide a function that directly manipulates `pluginManifest`.

```golang
var (
	pluginManifestLock sync.RWMutex
)

// Overrides the plugin manifest
// This function is used to override the plugin manifest for testing purposes etc...
func OverridesPluginManifest(pluginTestManifestJSON []byte) {
	pluginManifestLock.Lock()
	defer pluginManifestLock.Unlock()
	pluginManifest = manifest.MustParseSystemFromBytes(pluginTestManifestJSON, nil, pluginTranslationList.TranslatedRef())
}

// Restores the plugin manifest to the original manifest
func RestorePluginManifest() {
	pluginManifestLock.Lock()
	defer pluginManifestLock.Unlock()
	pluginManifest = manifest.MustParseSystemFromBytes(pluginManifestJSON, nil, pluginTranslationList.TranslatedRef())
}

func GetPropertySchemaByVisualizer(v visualizer.Visualizer) *property.Schema {
	pluginManifestLock.RLock()
	defer pluginManifestLock.RUnlock()

	for _, p := range pluginManifest.ExtensionSchema {
		if p.ID().String() == "reearth/"+string(v) {
			return p
		}
	}
	return nil
}
...
```

スクリーンショット 2025-03-31 18 33 29

By doing this, we should be able to write test code where needed and run tests without having to provide a function that directly manipulates pluginManifest.

```golang
      package e2e
      import (
	    "testing"

	     _ "embed"

	    "github.com/reearth/reearth/server/pkg/builtin"
	    "golang.org/x/text/language"
       )

       //go:embed test_data/manifest.yml
        var pluginTestManifestJSON []byte
        ....        
        ....

	builtin.OverridePluginManifest(pluginTestManifestJSON)
	defer builtin.RestorePluginManifest()

	// 1. tile_opacity
	// 2. tile_type
	// 3. tile_zoomLevel
	res = getScene(e, sId, language.Und.String())
	fields = res.Path("$.property.items[0].groups[0].fields").Array()
	fields.Value(0).Object().Value("fieldId").IsEqual("tile_opacity")
	fields.Value(1).Object().Value("fieldId").IsEqual("tile_type")
	fields.Value(2).Object().Value("fieldId").IsEqual("tile_zoomLevel")
```

soneda-yuya: I submitted [suggestion PR](https://github.com/reearth/reearth-visualizer/pull/1539) for this.

pyshx (approved):

PR #1539, opened 2025-03-31
chore(server): suggestion - revert sort logic [VIZ-1423]

soneda-yuya:
# Overview

### DIFF
--- a/server/e2e/test_data/manifest.yml
+++ b/server/e2e/test_data/manifest.yml
@@ -1,0 +1,81 @@
+version: ../../schemas/plugin_manifest.json
+id: reearth
+system: true
+name: Re:Earth Official Plugin
+description: Official Plugin
+author: Re:Earth
+extensions:
+  - id: cesium-beta
+    name: Cesium Beta
+    description: Description
+    visualizer: cesium
+    type: visualizer
+    schema:
+      groups:
+        - id: default
+          collection: main
+          title: Main
+          fields:
+            - id: ion
+              type: string
+              title: Cesium Ion API access token
+              description: Cesium Ion account users may use their personal API keys to be able to use their Cesium Ion assets(tile data, 3D data, etc) with their project.
+            - id: vr
+              type: bool
+              title: VR
+              description: Enable VR mode to split the screen into left and right.
+        - id: tiles
+          collection: tiles
+          title: Tiles
+          description: You may change the look of the Earth by obtaining map tile data and setting it here.
+          list: true
+          representativeField: tile_type
+          fields:       
+            - id: tile_type
+              type: string
+              title: Tile type
+              defaultValue: default
+              placeholder: please enter tile type
+              choices:
+                - key: default
+                  label: Default
+                - key: default_label
+                  label: Labelled
+                - key: default_road
+                  label: Road Map
+                - key: open_street_map
+                  label: OpenStreetMap
+                - key: esri_world_topo
+                  label: ESRI Topography
+                - key: black_marble
+                  label: Earth at night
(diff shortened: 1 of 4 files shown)


server/pkg/property/schema_group.go
```diff
@@ -118,29 +118,3 @@ func (s *SchemaGroup) RepresentativeField() *SchemaField {
...
-	// Remove the field from the original position
-	s.fields = append(s.fields[:from], s.fields[from+1:]...)
-
-	// Insert the field into the new position
-	if to >= len(s.fields) {
-		s.fields = append(s.fields, field)
-	} else {
-		s.fields = append(s.fields[:to], append([]*SchemaField{field}, s.fields[to:]...)...)
-	}
-}
```
soneda-yuya, line 146: I’d prefer to avoid operations that modify the schema itself.

server/e2e/test_data/manifest.yml
```diff
@@ -0,0 +1,81 @@
...
+              placeholder: please enter tile url
+            - id: tile_zoomLevel
+              type: array
+              title: Zoom level
+              description: "Change the zoom level of the selected tile map. Min:0, Max:25"
+              defaultValue: [0,25]
+              ui: zoomLevel
+              min: 0
+              max: 25
+              placeholder: please enter tile zoom level
```
soneda-yuya, line 81: We can prepare test data like this when we need it,

server/pkg/builtin/main.go
```diff
@@ -34,6 +35,13 @@ var (
 	PropertySchemaIDStoryBlock           = id.MustPropertySchemaID("reearth/storyBlock")
 )
 
+// Overrides the plugin manifest
+// Ideally, we’d like to avoid defining this function here,
+// but it was implemented as a necessary workaround for testing pluginManifest　 on internal pkg.
+func OverridePluginManifest(pluginTestManifestJSON []byte) {
+	pluginManifest = manifest.MustParseSystemFromBytes(pluginTestManifestJSON, nil, pluginTranslationList.TranslatedRef())
+}
+
```
soneda-yuya, line 44: As noted above, we’d prefer not to define this here,
but it's necessary for testing pluginManifest.
That said, `OverridePluginManifest` is generic and does not depend on any specific test case.

server/pkg/builtin/main.go
```diff
@@ -34,7 +40,27 @@ var (
...
+func OverridePluginManifest(pluginTestManifestJSON []byte) {
+	pluginManifestLock.Lock()
+	pluginManifest = manifest.MustParseSystemFromBytes(pluginTestManifestJSON, nil, pluginTranslationList.TranslatedRef())
+}
+
+// Restores the plugin manifest to the original manifest
+func RestorePluginManifest() {
+	defer pluginManifestLock.Unlock()
+	pluginManifest = manifest.MustParseSystemFromBytes(pluginManifestJSON, nil, pluginTranslationList.TranslatedRef())
+}
```
soneda-yuya, line 58: The pluginManifest is initialized when this package is loaded.
Therefore, if you modify the contents of pluginManifest in a test, make sure to restore it afterward — otherwise, it may affect other tests.

soneda-yuya: Since it’s a global variable, even if you use locks, it’s difficult to overwrite it only within a single test case.
Because it’s global, the same value is basically referenced throughout the tests.

Therefore, to fully isolate each test, you need to initialize it using something like a struct instead of a global variable.
However, since the Visualizer tests are generally executed sequentially and considering the cost involved, that level of handling has not been implemented.

PR #1586, opened 2025-04-24
fix(web): fix TS error in vite.config.ts file [VIZ-1559]

mulengawilfred:
# Overview
- Fixes TypeScript warning regarding rollup plugin, yaml, which had no matching type with that for vite plugins. Solution was to add a cast. 
- Moves test configuration logic to `vitest.config.ts`  
## Summary by CodeRabbit

- **Chores**
	- Updated test configuration to separate testing setup from the main configuration.
	- Adjusted plugin usage and test environment settings for improved test management.
	- Refined coverage reporting and module aliasing for tests.

### DIFF
--- a/web/vite.config.ts
+++ b/web/vite.config.ts
@@ -8,11 +8,10 @@ import { resolve } from "path";
 import yaml from "@rollup/plugin-yaml";
 import react from "@vitejs/plugin-react-swc";
 import { readEnv } from "read-env";
-import { defineConfig, loadEnv, type Plugin } from "vite";
+import { defineConfig, loadEnv, PluginOption, type Plugin } from "vite";
 import cesium from "vite-plugin-cesium";
 import svgr from "vite-plugin-svgr";
 import tsconfigPaths from "vite-tsconfig-paths";
-import { configDefaults } from "vitest/config";
 
 import pkg from "./package.json";
 
@@ -44,7 +43,7 @@ export default defineConfig({
   plugins: [
     svgr(),
     react(),
-    yaml(),
+    yaml() as PluginOption,
     cesium({
       cesiumBaseUrl: cesiumVersion ? `cesium-${cesiumVersion}/` : undefined
     }),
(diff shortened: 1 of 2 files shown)


web/vite.config.ts
```diff
@@ -44,7 +43,7 @@ export default defineConfig({
   plugins: [
     svgr(),
     react(),
-    yaml(),
+    yaml() as PluginOption,
```
airslice, line 46: Could you help provide the reference why we need this? Usually the type issue here might coming from mis-match of versions between plugin and vite, could you try solve it on this direction? Or if it's a known issue from this plugin please add comments with link to the issue. Thank you!

web/vitest.config.ts
```diff
@@ -1,6 +1,40 @@
...
+import { PluginOption } from "vite";
+import svgr from "vite-plugin-svgr";
+import tsconfigPaths from "vite-tsconfig-paths";
+import { defineConfig, mergeConfig, configDefaults } from "vitest/config";
 
-import viteConfig from "./vite.config.ts";
-
-// This file is required by VSCode's Vitest extension.
-export default mergeConfig(viteConfig, defineConfig({}));
+export default mergeConfig(
```
airslice, line 7: Why do we need mergeConfig here?
BTW how did you pick the plugins that will be needed here?

web/vite.config.ts
mulengawilfred: I've checked that this [plugin](https://www.npmjs.com/package/@rollup/plugin-yaml) was last updated 2 years ago so going the route of fixing dependencies based on versions might not work. I think the used `any` type in their plugin type definition that made it incompatible with that from vite. Casting here seems like the best option.

web/vitest.config.ts
mulengawilfred: The plugins are required when the tests run and are from vite.config.ts. I've updated the approach to merge to config from vite.config.ts so that I don't have to import the plugins in vitest.config.ts.

PR #1587, opened 2025-04-25
feat(server): update alias check [VIZ-1541]

hexaforce:
# Overview

## What I've done
* Update
```
query CheckProjectAlias($alias: String!, $projectId: ID) {
  checkProjectAlias(alias: $alias, projectId: $projectId) {
    alias
    available
  }
}
```
* New!
```
query CheckStoryAlias($alias: String!, $storyId: ID) {
  checkStoryAlias(alias: $alias, storyId: $storyId) {
    alias
    available
  }
}
```

### 2. These APIs determine whether the given alias is currently available.
If `available=false` is returned from this API, the alias generally cannot be used.

### 3. The alias parameter is required, but the ID is optional.
```
extend type Query {
  checkProjectAlias(alias: String!, projectId: ID)
  checkStoryAlias(alias: String!, storyId: ID)
}
```
Currently, there are two privileged aliases that only you can always acquire:
* Your own ID (self ID)
* Your own ID with the prefix (p- or s- + self ID)
#### In other words, if you do not specify an ID, the system cannot determine who you are, so a stricter check will be performed.
Note that the prefix is not added automatically, so if you want to check with a prefix, you need to include it (e.g. p-{projectId} or s-{storyId}).

## Summary by CodeRabbit

- **New Features**
  - Added the ability to check the availability of story aliases via a new GraphQL query.
  - Enhanced project alias availability checks with support for specifying a project ID.
- **Bug Fixes**
  - Improved validation and error handling for alias availability, including reserved prefixes and uniqueness across projects and stories.
- **Tests**
  - Introduced comprehensive end-to-end tests to verify alias availability logic for both projects and stories.
- **Documentation**
  - Updated GraphQL schema documentation to reflect new and updated queries for alias availability.


server/internal/usecase/interactor/project.go
```diff
@@ -653,23 +703,23 @@ func updateProjectUpdatedAtByScene(ctx context.Context, sceneID id.SceneID, r re
...
-		return true, nil
-	}
+// 	if prj.ID() == updatedProjectID {
+// 		return true, nil
+// 	}
 
-	return false, interfaces.ErrProjectAliasAlreadyUsed
-}
+// 	return false, interfaces.ErrProjectAliasAlreadyUsed
+// }
```
soneda-yuya, line 722: if we don't need these lines, could you delete them?

hexaforce: yes

soneda-yuya (approved):
LGTM!

PR #1591, opened 2025-05-08
fix(web): skip project mutation update when no change [VIZ-1523]

mkumbobeaty:
# Overview
- Skip update mutation when there’s no change been made in project settings
## Summary by CodeRabbit

- **New Features**
  - Input, textarea, and number fields now support an "onChangeComplete" event, allowing updates to trigger immediately after editing is finished rather than only on blur.

- **Improvements**
  - Streamlined form interactions by removing unnecessary local state in several settings and editor panels.
  - Enhanced input responsiveness and reduced redundant updates in asset, color, and number fields.
  - Improved consistency in when updates are applied across project, workspace, and editor settings.

- **Bug Fixes**
  - Prevented redundant or unnecessary update calls when field values have not changed.


web/src/beta/ui/fields/InputField/index.tsx
```diff
@@ -2,19 +2,47 @@ import { TextInput, TextInputProps } from "@reearth/beta/lib/reearth-ui";
...
     "value" | "placeholder" | "onChange" | "onBlur" | "disabled" | "appearance"
   >;
 
-const InputField: FC<InputFieldProps> = ({ title, description, ...props }) => {
+const InputField: FC<InputFieldProps> = ({
+  title,
+  description,
+  value,
+  onChange,
+  onBlur,
```
airslice, line 18: Since we added the logic of checking value, i feel `onBlur` is not a proper name now, could we rename it as `onChangeComplete` ? In this case
`onChange` -> trigger on every change
`onChangeComplete` -> trigger on blur and there do have changes
What do you think?
For fields we could remove `onBlur`, i guess no component wants blur exactly, all they need is chagneComplete, please let me know if you find any use case it needs `onBlur`, if so we can still have it.

airslice: Same on other fields in this PR.

airslice (approved):

PR #1611, opened 2025-05-20
feat(server): add healthcheck path

franzramadhan:
# Overview


server/internal/app/health.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"cloud.google.com/go/storage"
+	"github.com/hellofresh/health-go/v5"
+	"github.com/hellofresh/health-go/v5/checks/mongo"
+	"github.com/labstack/echo/v4"
+	"github.com/reearth/reearth/server/internal/app/config"
+	"github.com/reearth/reearthx/log"
+)
+
+// HealthCheck returns an echo.HandlerFunc that serves the health check endpoint
```
pyshx, line 19: Please remove this comment, function name and parameters are self-explanatory.

pyshx (commented):
Shall we also add Mongo check?

franzramadhan:
> Shall we also add Mongo check?

@pyshx this should be covered in https://github.com/reearth/reearth-visualizer/pull/1611/files#diff-2857901e3ce84aa30fa99bb3ddc0c4d89eb2ea5e0e748053ffdf3f16376c2f5dR29

server/go.mod
pyshx, line 1: It would be better if you can create a seperate PR for deps upgrade

franzramadhan: @pyshx it's resulted from `go mod tidy`

pyshx (approved):

PR #1646, opened 2025-06-04
fix(web): updating workspace name setting

mkumbobeaty:
# Overview


web/src/beta/features/WorkspaceSetting/innerPages/Workspaces/index.tsx
```diff
@@ -21,17 +21,17 @@ type Props = {
...
+
+  const handleSubmitUpdateWorkspaceName = useCallback(
+    async (name: string) => {
+      await updateWorkspace({
+        teamId: workspace?.id ?? "",
+        name
+      });
+    },
+    [workspace, updateWorkspace]
+  );
```
airslice: - There's no follow up actions, so maybe we don't need async await in this case.
- Better return directly if no workspace?.id instead of calling api with empty workspace id.

airslice (approved):

PR #1664, opened 2025-06-13
feat(web): add getCurrentLocation example in playground

asahina820:
# Overview
This PR adds getCurrentLocation example in playground

## What I've done
- Added a new plugin example that demonstrates getCurrentLocationAsync() API usage. The plugin allows users to retrieve their current location and fly the camera to that position with error handling for cases where location cannot be obtained.

image

## Summary by CodeRabbit

- **New Features**
  - Introduced a "Get Current Location" plugin in the Plugin Playground, enabling users to retrieve their current geographic location and fly the camera to that position within the viewer.
- **Localization**
  - Added English and Japanese translations for the "Get Current Location" plugin label.


web/src/beta/features/PluginPlayground/Plugins/presets/viewer/getCurrentLocation.ts
```diff
@@ -0,0 +1,137 @@
...
+    description: Get Current Location and Fly To
+  `,
+  disableEdit: true,
+  disableDelete: true
+};
+
+const widgetFile: FileType = {
+  id: "get-current-location",
+  title: "get-current-location.js",
+  sourceCode: `reearth.ui.show(\`
```
ShogoHirasawa: I thought it would be better to first describe what this sample shows.
e.g.  // This is sample code how to show user's current location

web/src/beta/features/PluginPlayground/Plugins/presets/viewer/getCurrentLocation.ts
```diff
@@ -0,0 +1,137 @@
...
+    });
+  </script>
+\`);
+
+// Handle button click to get current location
+// Documentation for Viewer "getCurrentLocationAsync" function https://visualizer.developer.reearth.io/plugin-api/viewer/#-getcurrentlocationasync
+reearth.extension.on("message", async (msg) => {
+  if (msg.type === "getCurrentLocation") {
+    const myLocation = await reearth.viewer.tools.getCurrentLocationAsync();
+
```
ShogoHirasawa, line 108: How about showing a marker on the Visualizer to indicate the user's current location.
Since the Plugin Playground is used a satellite image as base map, it may be difficult to notice that the current location. the marker help to indicate the current location would be kind for user.

asahina820: That sounds like a good idea, but it would require updating the marker position each time the location data is retrieved, which would complicate the entire code.
I thought that displaying and updating point data is in another example and could be implemented by the user. What do you think?

web/src/beta/features/PluginPlayground/Plugins/presets/viewer/getCurrentLocation.ts
asahina820: I added, thank you!

web/src/beta/features/PluginPlayground/Plugins/presets/viewer/getCurrentLocation.ts
asahina820: Resolved by discussion. No additional markers were added.

ShogoHirasawa (approved):
LGTM! Thank you!

PR #1677, opened 2025-06-25
test(web): fix playwright config and login test

waqaszulfiqar:
# Overview

### DIFF
--- a/web/e2e/tests/login.spec.ts
+++ b/web/e2e/tests/login.spec.ts
@@ -30,6 +30,7 @@ test.describe("Login Page Tests", () => {
     loginPage = new LoginPage(page);
     dashBoardPage = new DashBoardPage(page);
     await page.goto(REEARTH_WEB_E2E_BASEURL, { waitUntil: "networkidle" });
+    console.log(`Navigated to ${REEARTH_WEB_E2E_BASEURL}`);
   });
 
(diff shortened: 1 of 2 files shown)


web/e2e/tests/login.spec.ts
```diff
@@ -30,6 +30,9 @@ test.describe("Login Page Tests", () => {
     loginPage = new LoginPage(page);
     dashBoardPage = new DashBoardPage(page);
     await page.goto(REEARTH_WEB_E2E_BASEURL, { waitUntil: "networkidle" });
+    console.log(`Navigated to ${REEARTH_WEB_E2E_BASEURL}`);
+    console.log("username:", REEARTH_E2E_EMAIL);
+    console.log("password:", REEARTH_E2E_PASSWORD);
```
airslice: This will leads to password expose on log right? I think better avoid that.

airslice: We kept the account as secret by purpose, maybe remove the log for both username and password.
We might need to think about other solution for check login status.

airslice (approved):

wilfredmulenga (approved):
Looks good. 

Small request: Lets make sure PR titles and descriptions reflect the changes being made or include a bit of context about the impact.

PR #1691, opened 2025-07-04
feat(server): update internal project api [VIZ-1962,VIZ-1961]

hexaforce:
# Overview

### 1. Added a new internal API: PublishProject
This API is identical to the GraphQL API.
```diff
+rpc PublishProject(PublishProjectRequest) returns (PublishProjectResponse) {}

+message PublishProjectRequest {
+  // Project ID (required)
+  string project_id = 1;
+  // Scene publication alias
+  optional string alias = 2;
+  // Scene publication status
+  PublishmentStatus publishment_status = 3;
+}
```
### 2. Updated GetProjectList
* When the workspace ID is not specified, it retrieves all projects from workspaces that the user either owns (owner) or belongs to (member).
Note: If the user is only a workspace member, projects marked as deleted will not be retrieved.

* Added options for sorting, keyword search, and pagination.
A dedicated sorting logic has been newly implemented instead of using reearthx.
Unlike the GraphQL GetProject, this implementation fixes a bug that caused an infinite loop when searching in descending order.

```diff
message GetProjectListRequest {
  // Workspace ID
-  string workspace_id = 1;
+  optional string workspace_id = 1;
  // Authenticated flag
  bool authenticated = 2;
  // Keyword search
+  optional string keyword = 3;
+  // Sort options
+  optional ProjectSort sort = 4;
+  // Pagination information
+  optional Pagination pagination = 5;
}
```

### DIFF
--- a/server/internal/infrastructure/memory/project.go
+++ b/server/internal/infrastructure/memory/project.go
@@ -84,6 +84,10 @@ func (r *Project) FindByWorkspace(ctx context.Context, id accountdomain.Workspac
 	), nil
 }
 
+func (r *Project) FindByWorkspaces(ctx context.Context, authenticated bool, pFilter repo.ProjectFilter, owningWorkspaces accountdomain.WorkspaceIDList, wList accountdomain.WorkspaceIDList) ([]*project.Project, *usecasex.PageInfo, error) {
+	return nil, nil, nil
+}
+
 func (r *Project) FindStarredByWorkspace(ctx context.Context, id accountdomain.WorkspaceID) ([]*project.Project, error) {
 	r.lock.Lock()
 	defer r.lock.Unlock()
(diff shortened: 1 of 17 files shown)


server/internal/infrastructure/memory/project.go
```diff
@@ -146,40 +150,40 @@ func (r *Project) FindActiveByAlias(ctx context.Context, alias string) (*project
...
+// 			}
+// 		}
+// 	}
+
+// 	sort.Slice(result, func(i, j int) bool {
+// 		return result[i].UpdatedAt().After(result[j].UpdatedAt())
+// 	})
+
+// 	return result, nil
+// }
```
soneda-yuya: Could you remove if you use this code anymore.

server/pkg/project/sort_type.go
```diff
@@ -1,27 +1,25 @@
...
+// 	case "id":
+// 		return SortType{Key: "id", Desc: desc}
+// 	case "updatedat":
+// 		return SortType{Key: "updatedat", Desc: desc}
+// 	case "name":
+// 		return SortType{Key: "name", Desc: desc}
+// 	default:
+// 		return SortType{Key: "id", Desc: desc} // Default to ID sorting
+// 	}
+// }
```
soneda-yuya: Could you remove if you use this code anymore.

server/internal/infrastructure/memory/project.go
hexaforce: yes

server/pkg/project/sort_type.go
hexaforce: yes

PR #1698, opened 2025-07-09
feat(server): enable to access internal api to get project data without auth

soneda-yuya:
# Overview
- we need to get public project data without auth

## What I've done
- allow access to "GET" method on internal api
- update FindActiveById and FindVisibilityByWorkspace


server/internal/usecase/interactor/project.go
hexaforce, line 1: owingWsList -> owningWsLis

hexaforce (changes requested):
Visibility constantization
AccessMed check
is fine

server/internal/usecase/interactor/project.go
soneda-yuya: @hexaforce 
Thanks! I fixed!

hexaforce (approved):

PR #1739, opened 2025-07-25
feat(server): Project Data Maintenance

hexaforce:
# Overview

# Project Data Maintenance
## What I've done

* There are 9 projects without a projectAlias. These will be updated with default values starting with p-xxx.

* There are 2 project metadata records without an assigned workspace. These will be updated using the workspace associated with their respective projects.


soneda-yuya (approved):
LGTM!

PR #1748, opened 2025-07-28
fix(server): Cannot access a workspace

hexaforce:
# Overview

# Cannot access a workspace you do not belong to.

## What I've done
# The workspace was being restricted beyond just the query conditions.


zombozo12 (approved):
LGTM

PR #1750, opened 2025-07-28
fix(server): fix project filtering bug

hexaforce:
# Overview

# There was a mistake in the project filtering logic.

## 1.We will allow access to data from other workspaces via the internal API under the following conditions:
* The project's visibility is set to public
* The project is not in the trash


soneda-yuya (approved):
LGTM!

PR #1753, opened 2025-07-30
feat(server): add workspace alias [VIZ-2081]

hexaforce:
# Overview

# Add alias to the workspace API

## 1. The alias field is added to both workspace creation and update.
```diff
input CreateWorkspaceInput {
    name: String!
+    alias: String
}

input UpdateWorkspaceInput {
    workspaceId: ID!
    name: String!
+    alias: String
}
```
## 2.A uniqueness check will be performed when registering an alias.
```
CheckWorkspaceAliasUnique(context.Context, workspace.ID, string) error
```
## 3.If no alias is specified, it will be set to `w-{WorkspaceID}` by default.
```
	newAlias := fmt.Sprintf("w-%s", newId.String())
	if alias != nil {
		err := i.repos.Workspace.CheckWorkspaceAliasUnique(ctx, newId, *alias)
		if err != nil {
			return nil, err
		}
		newAlias = *alias
	}
```


soneda-yuya (approved):
LGTM!

PR #1754, opened 2025-07-31
feat(server): Remove filtering for internal API

hexaforce:
# Overview

# Remove filtering for internal API


soneda-yuya (approved):
LGTM!

PR #1757, opened 2025-08-05
feat(server): adjust visibility handling [VIZ-2110]

hexaforce:
# Overview

### 1. Added the following field to the exported project data:
* visibility

### 2. createProject

If the specified visibility is not allowed by the policy check, an error will be returned.

### 3. updateProject

If the specified visibility is not allowed by the policy check, an error will be returned.

### 4. importProject

If the imported data has visibility set to private and it is not allowed, the system will attempt to create the project with public visibility.

If the visibility is public and not allowed, the system will attempt to create the project with private visibility instead.

### DIFF
--- a/server/internal/usecase/interactor/project.go
+++ b/server/internal/usecase/interactor/project.go
@@ -311,30 +311,25 @@ func (i *Project) getMetadata(ctx context.Context, pList []*project.Project) ([]
 
 func (i *Project) Create(ctx context.Context, input interfaces.CreateProjectParam, operator *usecase.Operator, isImport bool) (_ *project.Project, err error) {
 
-	// Default to VisibilityPrivate if not specified
-	visibility := project.VisibilityPrivate
+	// Default to VisibilityPublic if not specified
+	visibility := project.VisibilityPublic
 	if input.Visibility != nil {
 		visibility = project.Visibility(*input.Visibility)
 	}
 
 	if i.policyChecker != nil {
 
-		if err = i.checkGeneralPolicy(ctx, input.WorkspaceID, visibility); err != nil {
-			// If this is an import, try the opposite visibility as a fallback
-			if isImport {
-				switch visibility {
-				case project.VisibilityPrivate:
-					visibility = project.VisibilityPublic
-				case project.VisibilityPublic:
-					visibility = project.VisibilityPrivate
-				default:
-					return nil, err
-				}
-				// Retry policy check with the new visibility
-				if err = i.checkGeneralPolicy(ctx, input.WorkspaceID, visibility); err != nil {
-					return nil, err
-				}
+		if isImport {
+			// Try checking if user can create private project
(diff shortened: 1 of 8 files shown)


server/internal/usecase/interactor/project.go
```diff
@@ -305,14 +309,46 @@ func (i *Project) getMetadata(ctx context.Context, pList []*project.Project) ([]
 	return pList, err
 }
 
-func (i *Project) Create(ctx context.Context, input interfaces.CreateProjectParam, operator *usecase.Operator) (_ *project.Project, err error) {
+func (i *Project) Create(ctx context.Context, input interfaces.CreateProjectParam, operator *usecase.Operator, isImport bool) (_ *project.Project, err error) {
+
+	// Default to VisibilityPrivate if not specified
+	visibility := project.VisibilityPrivate
```
soneda-yuya: @hexaforce 
Could you change default is public?

server/internal/usecase/interactor/project.go
```diff
@@ -305,14 +309,46 @@ func (i *Project) getMetadata(ctx context.Context, pList []*project.Project) ([]
...
+					return nil, err
+				}
+				// Retry policy check with the new visibility
+				if err = i.checkGeneralPolicy(ctx, input.WorkspaceID, visibility); err != nil {
+					return nil, err
+				}
+			} else {
+				return nil, err
+			}
+		}
```
soneda-yuya, line 335: Sorry, I mistook sentence :pray:

https://www.notion.so/eukarya/Sprint-64-23d16e0fb16580d18352edcda297daa1?source=copy_link#24216e0fb16580689949fbe062575cb5

**correct requirement**
- when user can’t create private project, import project visibility as public
- when user can create private project, import project visibility as private

```golang
if isImport {
	// Try checking if user can create private project
	errPrivate := i.checkGeneralPolicy(ctx, input.WorkspaceID, project.VisibilityPrivate)
	if errPrivate == nil {
		visibility = project.VisibilityPrivate
	} else {
		visibility = project.VisibilityPublic
	}
} else {
	// Not import: use given visibility
	if err = i.checkGeneralPolicy(ctx, input.WorkspaceID, visibility); err != nil {
		return nil, err
	}
}
```

server/internal/usecase/interactor/project.go
```diff
@@ -1106,26 +1163,26 @@ func (i *Project) createProject(ctx context.Context, input createProjectInput, o
...
+	// 	}
 
-		if err := p.EnforceProjectCount(projectCount + 1); err != nil {
-			return nil, err
-		}
-	}
+	// 	if err := p.EnforceProjectCount(projectCount + 1); err != nil {
+	// 		return nil, err
+	// 	}
+	// }
```
pyshx: This can be removed?

soneda-yuya (approved):
LGTM!

PR #1765, opened 2025-08-19
fix(web, server): correct data attribution[VIZ-2137]

ZTongci:
# Overview
Correct data attribution etc.

### DIFF
--- a/server/internal/infrastructure/mongo/migration/250820113009_change_esri_to_default.go
+++ b/server/internal/infrastructure/mongo/migration/250820113009_change_esri_to_default.go
@@ -2,83 +2,59 @@ package migration
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"time"
 
 	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
 func ChangeEsriToDefault(ctx context.Context, c DBClient) error {
-	collection := c.WithCollection("property").Client()
+	col := c.WithCollection("property").Client()
 
 	filter := bson.M{
-		"items.groups.fields": bson.M{
-			"$elemMatch": bson.M{
-				"field": "tile_type",
-				"value": "esri_world_topo",
-			},
+		"items.groups.fields.field": "tile_type",
+		"items.groups.fields.value": "esri_world_topo",
+	}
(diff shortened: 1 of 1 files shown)


web/src/app/features/Visualizer/Crust/Widgets/Widget/builtin/DataAttribution/index.tsx
```diff
@@ -57,6 +57,11 @@ const DataAttribution = ({
           <img src={cesiumCredit.logo} title={cesiumCredit.description} />
         </CesiumLink>
       )}
+      {googleCredit && (
+        <GoogleLink target="_blank" rel="noreferrer">
```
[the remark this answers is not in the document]
ZTongci: googleCredit.creditUrl now is undefined, add href will navigate to black page

server/internal/infrastructure/mongo/migration/250820113009_change_esri_to_default.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+	}
+
+	if err := cursor.Err(); err != nil {
+		return fmt.Errorf("cursor error: %w", err)
+	}
+
+	fmt.Printf("Migration completed. Updated %d documents.\n", updatedCount)
+	return nil
+}
```
soneda-yuya, line 60: I feel this is a bit risky since it overwrites the entire items.
If it happens to coincide with some user update at an unfortunate timing, there’s a possibility of overwriting their changes.
Would it be difficult to update only the necessary field instead?

soneda-yuya: FYI: I asked AI
```
"$set": bson.M{
  "items.$[].groups.$[].fields.$[f].value": "default",
}
```

	•	This performs a partial update on only the specific elements in the items array that match the condition.
	•	MongoDB internally rewrites only the value field of the matching array elements.
	•	Other fields elements or the items array as a whole are not affected.
	•	As a result, it is still an operation that updates items, but the update unit is not the entire array—only the specific field within the document’s nested elements.

soneda-yuya: https://github.com/reearth/reearth-visualizer/pull/1783

soneda-yuya (approved):
LGTM!

PR #1774, opened 2025-08-21
fix(server): policy check enum

soneda-yuya:
# Overview
remove unused policy check type and fix policy check type when user upload asset


server/internal/infrastructure/policy/permissive_checker_test.go
```diff
@@ -74,12 +74,12 @@ func TestPermissiveChecker_CheckPolicy(t *testing.T) {
 			name: "model count per project check",
```
pyshx: This doesn't match with `gateway.PolicyCheckCustomDomainCount`

pyshx (approved):

PR #1778, opened 2025-08-25
fix(server): import project data don't set visibility correctly

soneda-yuya:
# Overview
- import project -> x 
   - private workspace
       - private p import -> ok
       - public p import -> ok
   - public workspace
       - **private p import -> x => need to fix this case**
           - need to become public project but become private
       - public p import -> ok

### DIFF
--- a/server/internal/usecase/interactor/project.go
+++ b/server/internal/usecase/interactor/project.go
@@ -336,11 +336,8 @@ func (i *Project) getMetadata(ctx context.Context, pList []*project.Project) ([]
 }
 
 func (i *Project) Create(ctx context.Context, input interfaces.CreateProjectParam, operator *usecase.Operator) (_ *project.Project, err error) {
-	// Default to VisibilityPublic if not specified
 	visibility := project.VisibilityPublic
 
-	// if policy checker is not nil, check if the user can create a private project
-	// if user don't have permission to create a private project, set visibility to public
 	if i.policyChecker != nil {
 		errPrivate := i.checkGeneralPolicy(ctx, input.WorkspaceID, project.VisibilityPrivate)
 		if errPrivate != nil {
--- a/server/internal/usecase/interactor/project.go
+++ b/server/internal/usecase/interactor/project.go
@@ -335,35 +335,21 @@ func (i *Project) getMetadata(ctx context.Context, pList []*project.Project) ([]
 	return pList, err
 }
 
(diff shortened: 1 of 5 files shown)


server/internal/usecase/interactor/project.go
```diff
@@ -335,35 +335,21 @@ func (i *Project) getMetadata(ctx context.Context, pList []*project.Project) ([]
...
-			}
+		errPrivate := i.checkGeneralPolicy(ctx, input.WorkspaceID, project.VisibilityPrivate)
+		if errPrivate != nil {
+			visibility = project.VisibilityPublic
 		} else {
-			// Not import: use given visibility
-			if err = i.checkGeneralPolicy(ctx, input.WorkspaceID, visibility); err != nil {
-				return nil, err
+			if input.Visibility != nil {
+				visibility = project.Visibility(*input.Visibility)
```
soneda-yuya, line 347: When workspace don't have permission, system set visibility to public regardless of　create or import.

server/internal/usecase/interactor/project.go
```diff
@@ -1099,9 +1085,16 @@ func (i *Project) ImportProjectData(ctx context.Context, workspace string, proje
...
+	if i.policyChecker != nil {
+		errPrivate := i.checkGeneralPolicy(ctx, workspaceId, project.VisibilityPrivate)
+		if errPrivate == nil {
+			if ret, ok := projectData["visibility"].(string); ok {
+				vis := project.Visibility(ret)
+				visibility = vis
+			}
+		} else {
+			visibility = project.VisibilityPublic
+		}
```
soneda-yuya, line 1094: When workspace has permission, dist project inherit src project visibility.

server/internal/usecase/interactor/project.go
```diff
@@ -335,35 +335,21 @@ func (i *Project) getMetadata(ctx context.Context, pList []*project.Project) ([]
 	return pList, err
 }
 
-func (i *Project) Create(ctx context.Context, input interfaces.CreateProjectParam, operator *usecase.Operator, isImport bool) (_ *project.Project, err error) {
-
+func (i *Project) Create(ctx context.Context, input interfaces.CreateProjectParam, operator *usecase.Operator) (_ *project.Project, err error) {
```
soneda-yuya, line 338: When user import project, actually system call createProject twice on Create and ImportProjectData function.
In addition visibility is checked on ImportProjectData function, so I delete isImport flag from here.

PR #1825, opened 2025-09-24
feat(server): Allow access from storage trigger [VIZ-2232]

hexaforce:
# Overview

# Allow access from storage trigger

## {workspaceID}-{projectID}-{userID}.zip
```go
func GenFileName(workspaceID accountdomain.WorkspaceID, projectID id.ProjectID, userID accountdomain.UserID) string {
	return fmt.Sprintf("%s-%s-%s.zip", workspaceID.String(), projectID.String(), userID.String())
}
```

### 2.When access comes from the storage trigger, the operator is supplemented:
```go
return func(c echo.Context) error {
    req := c.Request()
    ctx := req.Context()
    ...
    if req.Method == "POST" && req.URL.Path == "/api/import-project" {
        _, _, _, userID, err := SplitFilename(n.CloudEventData.Name)
        ...
        u, err := cfg.AccountRepos.User.FindByID(ctx, *userID)
        ...
        op, err := generateOperator(ctx, cfg, u)
        ...
        ctx = adapter.AttachOperator(ctx, op)
```

### DIFF
--- a/server/internal/app/app.go
+++ b/server/internal/app/app.go
@@ -153,8 +153,8 @@ func initEcho(ctx context.Context, cfg *ServerConfig) *echo.Echo {
 
 	serveFiles(e, allowedOrigins(cfg), cfg.Gateways.DomainChecker, cfg.Gateways.File)
 
-	servSplitUploadFiles(e, cfg.Gateways.File)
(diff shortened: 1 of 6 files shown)


soneda-yuya (approved):
LGTM!

PR #1827, opened 2025-09-25
fix(server): delete validation of url geojson

soneda-yuya:
# Overview
I removed the GeoJSON input validation from external URLs.
The reason is that since this is an external resource, our backend does not need to guarantee the correctness of its format.
Now we only check the validity of the URL itself, while the GeoJSON format validation is handled on the frontend.


wilfredmulenga (approved):
Looks good

PR #1828, opened 2025-09-25
fix(web, server): fixed typo import result log

hexaforce:
# Overview

## fixed typo import result log
```
imporResultLog -> importResultLog
```


soneda-yuya (approved):
LGTM!

PR #1845, opened 2025-10-08
fix(server): fix pagination for GetAllProjects to use limit and offset param [VIZ-2271]

wilfredmulenga:
- As per PR title

### DIFF
--- a/server/internal/infrastructure/mongo/project.go
+++ b/server/internal/infrastructure/mongo/project.go
@@ -606,8 +606,61 @@ func (r *Project) FindAll(ctx context.Context, pFilter repo.ProjectFilter) ([]*p
 		}
 	}
 
+	if pFilter.Limit != nil && pFilter.Offset != nil {
+		
+		totalCount, err := r.client.Count(ctx, filter)
+		if err != nil {
+			log.Errorf("FindAll: Count error: %v", err)
+			return nil, nil, rerror.ErrInternalByWithContext(ctx, err)
+		}
+
+		var sortDoc bson.D
+		if pFilter.Sort != nil {
+			sortKey := pFilter.Sort.Key
+			if sortKey == "starcount" {
+				sortKey = "star_count" // Map to MongoDB field name
+			}
+			sortOrder := 1
+			if pFilter.Sort.Desc {
+				sortOrder = -1
+			}
+			sortDoc = bson.D{{Key: sortKey, Value: sortOrder}}
+		}
+
+		collation := options.Collation{
+			Locale:    "en",
+			Strength:  3,
+			CaseLevel: true,
+			Alternate: "shifted",
+		}
+
+		findOptions := options.Find().
+			SetCollation(&collation).
+			SetSkip(*pFilter.Offset).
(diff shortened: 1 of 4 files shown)


hexaforce (approved):

server/internal/infrastructure/mongo/project.go
```diff
@@ -606,8 +606,61 @@ func (r *Project) FindAll(ctx context.Context, pFilter repo.ProjectFilter) ([]*p
...
+
+		pageInfo := &usecasex.PageInfo{
+			TotalCount:      totalCount,
+			HasNextPage:     *pFilter.Offset+*pFilter.Limit < totalCount,
+			HasPreviousPage: *pFilter.Offset > 0,
+		}
+
+		return c.Result, pageInfo, nil
+	}
+
```
soneda-yuya, line 662: @wilfredmulenga 
When you return error, could you use `visualizer.ErrorWithCallerLogging`.

server/internal/usecase/interfaces/project.go
```diff
@@ -93,6 +93,11 @@ type ProjectListParam struct {
 	Offset *int64
 }
 
+type AllProjectsParam struct {
+	Limit  *int64
+	Offset *int64
+}
+
```
soneda-yuya: @wilfredmulenga 
Can we use `ProjectListParam` which is already defined.

I think it’s difficult to tell the difference between `ProjectListParam` and `AllProjectsParam` just from their names.

If the idea behind this is that parameters like topic or other search filters might be added later, then I think that logic could be shared and reused for other types of searches as well.

Even if data retrieval is separated at the repository level, I feel it would be better to shape it into a consistent form that follows the repository interface within the use case layer.

That said, this is just my personal opinion — I’ll leave the final decision to you.

server/internal/infrastructure/mongo/project.go
wilfredmulenga: done

server/internal/usecase/interfaces/project.go
wilfredmulenga: done

server/internal/infrastructure/mongo/project.go
```diff
@@ -606,8 +607,61 @@ func (r *Project) FindAll(ctx context.Context, pFilter repo.ProjectFilter) ([]*p
...
+			HasNextPage:     *pFilter.Offset+*pFilter.Limit < totalCount,
+			HasPreviousPage: *pFilter.Offset > 0,
+		}
+
+		return c.Result, pageInfo, visualizer.ErrorWithCallerLogging(ctx, "FindAll: Count error:", err)
+	}
+
 	if pFilter.Pagination == nil {
-		log.Warnf("FindAll: Pagination is nil, not using pagination")
+		log.Warnf("FindAll: No pagination provided, returning all results")
```
soneda-yuya: @wilfredmulenga 
Could you use `visualizer.WarnWithCallerLogging`?

wilfredmulenga: done

PR #2178, opened 2026-04-24
refactor(web): improve plugin playground content sharing

mkumbobeaty:
# Overview
Sharing plugins in the plugin playground can fail with a 414 (URI Too Long) error when the plugin content is too large. This happens because the current sharing mechanism encodes all plugin data directly into the URL, which exceeds length limits imposed by browsers and servers.

To address this, sharing is now blocked when content exceeds safe URL limits. Users are shown a popup alert and guided to use the export function as an alternative for handling large plugins.

## What I've done
- Check content length before sharing
- Block if it exceeds the limit 
- Show popup modal with guidance to export plugin

Screenshot 2026-04-24 at 11 11 33

## Checklist


web/src/app/features/PluginPlayground/Plugins/constants.ts
```diff
@@ -14,3 +14,4 @@ export type FileType = {
 export const ALLOWED_FILE_EXTENSIONS = ["js"] as const;
 
 export const SHARED_PLUGIN_ID = "shared-plugin-id";
+export const MAX_SHARE_URL_LENGTH = 5000;
```
airslice: Could you add a comment explaining why this limit exists and how the value 5000 was determined?

airslice (approved):

PR #2180, opened 2026-04-24
fix(server): batch project asset deletion

wilfredmulenga:
Fixes: [SCA-04] [HIGH] Unbounded asset load + N+1 deletes during project deletion

`RemoveByProjectWithFile` loaded every asset for a project into memory and deleted them serially (GCS + DB per asset). Projects with 200+ assets already hit Cloud Run timeouts in prod logs. Now paginates in batches of 100, parallelises GCS deletes (10 at a time), and uses `DeleteMany` per batch.

Also adds basic mongo-level tests for `RemoveByProjectWithFile`, `FindByIDs`, and `Remove`, which had no coverage.

Reported in: https://github.com/eukarya-inc/compliance/issues/7 (SCA-04)


server/internal/infrastructure/mongo/asset.go
```diff
@@ -160,39 +171,61 @@ func (r *Asset) Remove(ctx context.Context, id id.AssetID) error {
...
+		for _, a := range batch {
+			a := a
+			g.Go(func() error {
+				u, perr := url.Parse(a.URL())
+				if perr != nil || u == nil {
+					return nil
+				}
+				if err := f.RemoveAsset(gctx, u); err != nil {
+					// Preserve existing behavior: log and continue. The DB row
+					// is still removed below so a retry can re-delete it.
```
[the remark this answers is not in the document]
wilfredmulenga: fixed, will preserve db row

server/internal/infrastructure/mongo/asset.go
```diff
@@ -160,39 +171,69 @@ func (r *Asset) Remove(ctx context.Context, id id.AssetID) error {
...
-		if err != nil {
-			log.Print(err.Error())
+		var (
+			mu      sync.Mutex
+			failIDs = map[string]struct{}{}
+		)
+		g, gctx := errgroup.WithContext(ctx)
+		g.SetLimit(removeByProjectMaxConcurrent)
+		for _, a := range batch {
+			a := a
```
soneda-yuya: @wilfredmulenga 
I think we don't need this assignment after go version 1.22

server/internal/infrastructure/mongo/asset.go
```diff
@@ -160,39 +171,69 @@ func (r *Asset) Remove(ctx context.Context, id id.AssetID) error {
...
+			}
+		}
+		writeFilter := applyWorkspaceFilter(bson.M{"id": bson.M{"$in": ids}}, r.f.Writable)
+		if err := r.client.RemoveAll(ctx, writeFilter); err != nil {
+			return err
 		}
 
+		if len(batch) < removeByProjectBatchSize {
+			return nil
+		}
```
soneda-yuya, line 227: @wilfredmulenga 
This process can result in an infinite loop.

The loop repeatedly fetches a batch using the same baseFilter:

```go
batch, err := r.find(ctx, baseFilter, options.Find().SetLimit(removeByProjectBatchSize))
```

It relies on successful records being removed from the DB so that the next find call naturally returns the next batch.

However, if file/GCS deletion fails, that record is intentionally excluded from DB deletion so that it can be retried later:

```go
// Log and skip DB deletion so the row survives for retry.
failIDs[a.ID().String()] = struct{}{}
```

Because of that, failed records remain in the DB and still match baseFilter.

If all records in a batch fail to be deleted from storage, then ids becomes empty, no DB rows are removed, and the DB state does not change. The next loop iteration will fetch the same batch again, causing the same failures repeatedly.

wilfredmulenga: Good catch. If all records in a batch fail GCS deletion the loop will spin forever on the same batch. One way to fix this would be to always delete the DB row regardless of GCS failure, accepting that failed GCS objects become leftover, and rely on GCS lifecycle rules as a backstop for cleanup. Are we okay with leftover GCS objects in that failure scenario, or do we need a stricter cleanup guarantee?

server/internal/infrastructure/mongo/asset.go
wilfredmulenga: fixed

server/internal/infrastructure/mongo/asset.go
soneda-yuya: I think that approach is fine.

However, we should make sure to leave an error log when GCS deletion fails, so we can still detect and investigate leftover objects if needed.

wilfredmulenga: Sorry for the delay. I fixed it in this commit 6ae9b541a691b9ac795266c3c3e9ec87953338f6

server/internal/infrastructure/mongo/asset.go
```diff
@@ -160,39 +170,61 @@ func (r *Asset) Remove(ctx context.Context, id id.AssetID) error {
...
-		err = f.RemoveAsset(ctx, aPath)
-		if err != nil {
-			log.Print(err.Error())
+		g, gctx := errgroup.WithContext(ctx)
+		g.SetLimit(removeByProjectMaxConcurrent)
+		for _, a := range batch {
+			g.Go(func() error {
+				u, perr := url.Parse(a.URL())
+				if perr != nil || u == nil {
+					return nil
```
soneda-yuya, line 203: @wilfredmulenga 
Should we put warning log or something here?

soneda-yuya (approved):
LGTM!

server/internal/infrastructure/mongo/asset.go
```diff
@@ -160,39 +170,61 @@ func (r *Asset) Remove(ctx context.Context, id id.AssetID) error {
...
+				if err := f.RemoveAsset(gctx, u); err != nil {
+					// GCS delete failed; log for investigation. The DB row is
+					// still removed below so the loop terminates — orphaned
+					// objects must be cleaned up via GCS lifecycle rules.
+					log.Errorfc(gctx, "asset: gcs delete failed for %s: %v", a.ID(), err)
+				}
+				return nil
+			})
 		}
+		_ = g.Wait()
```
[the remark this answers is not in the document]
wilfredmulenga: Already discussed above

server/internal/infrastructure/mongo/asset.go
wilfredmulenga: Added a warning log in the latest commit.

wilfredmulenga (commented):
Looks good

soneda-yuya (approved):

PR #2185, opened 2026-05-08
fix(server): bound and pre-filter FindAll aggregation pipelines

wilfredmulenga:
Fixes: [SCA-03] [HIGH] MongoDB \$lookup joins entire collection before pagination in FindAll aggregation

`findAllWithTopicsFilter` and `findAllWithStarcountSort` had no limit when `pFilter.Limit`/`pFilter.Offset` were nil, causing unbounded joins against `projectmetadata`. Fixed by always applying pagination (default 100, hard cap 200) and moving the topics filter inside a pipeline-style `\$lookup`.

The Copilot comments on cursor pagination / negative values are minor — `GetAllProjects` (the only caller) already caps at 100 and converts cursor params to explicit limit/offset before reaching this code. Tested with live data.

Reported in: https://github.com/eukarya-inc/compliance/issues/7 (SCA-03)

- [x] Verified backward compatibility related to feature modifications.
- [x] Verified that no PII is included in any displayed values.


server/internal/infrastructure/mongo/project.go
```diff
@@ -31,6 +31,32 @@ var (
...
+// default when nil and capping at findAllMaxLimit.
+func resolveFindAllPagination(pFilter repo.ProjectFilter) (offset, limit int64) {
+	if pFilter.Offset != nil {
+		offset = *pFilter.Offset
+	}
+	if pFilter.Limit != nil {
+		limit = *pFilter.Limit
+	} else {
+		limit = findAllDefaultLimit
+	}
```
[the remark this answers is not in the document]
wilfredmulenga: Verified locally — the only caller of `FindAll` is `GetAllProjects` in the gRPC internal API (`server.go`), which always applies its own default of 100 and a cap of 100 before calling the interactor. Negative values are not reachable from this path.

On cursor pagination: the gRPC layer converts cursor `First`/`Last` into `param.Limit`/`param.Offset` (explicit fields), so `pFilter.Pagination.Cursor` is never populated when going through this code path. `resolveFindAllPagination` reading `pFilter.Limit` is sufficient.

Tested both the starcount-sort and topics-filter aggregation paths with live data — results are correctly bounded and filtered.

server/internal/infrastructure/mongo/project.go
```diff
@@ -1006,11 +1038,10 @@ func (r *Project) findAllWithTopicsFilter(ctx context.Context, pFilter repo.Proj
...
 
-	// Handle pagination
-	if pFilter.Limit != nil && pFilter.Offset != nil {
-		pipeline = append(pipeline, bson.M{"$skip": *pFilter.Offset})
-		pipeline = append(pipeline, bson.M{"$limit": *pFilter.Limit})
-	}
+	// Always paginate to keep the joined result set bounded.
+	offset, limit := resolveFindAllPagination(pFilter)
+	pipeline = append(pipeline, bson.M{"$skip": offset})
+	pipeline = append(pipeline, bson.M{"$limit": limit})
```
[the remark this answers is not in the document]
wilfredmulenga: Same as above — cursor pagination is handled at the gRPC layer, which maps `First`/`Last` to `param.Limit`/`param.Offset` before `FindAll` is called. By the time the request reaches the aggregation pipeline, `pFilter.Limit` is always set. No cursor conditions are needed in the aggregation path.

server/internal/infrastructure/mongo/project.go
```diff
@@ -1006,11 +1038,10 @@ func (r *Project) findAllWithTopicsFilter(ctx context.Context, pFilter repo.Proj
 		}
 	}
 
-	// Handle pagination
-	if pFilter.Limit != nil && pFilter.Offset != nil {
-		pipeline = append(pipeline, bson.M{"$skip": *pFilter.Offset})
-		pipeline = append(pipeline, bson.M{"$limit": *pFilter.Limit})
-	}
+	// Always paginate to keep the joined result set bounded.
+	offset, limit := resolveFindAllPagination(pFilter)
```
soneda-yuya: @wilfredmulenga 
Are there any remaining implementations, including on the frontend side, that still assume pFilter.Limit and pFilter.Offset can be nil?

wilfredmulenga: Thanks Yuya-san. I have confirmed that the only caller of FindAll is GetAllProjects which is used in gRPC and has a default limit and offset. Its called by dashboard. If dashboard does not provide a pagination value, then default will be used by visualizer which is 100 for limit and 0 for offset.

soneda-yuya: >If dashboard does not provide a pagination value, then default will be used by visualizer which is 100 for limit and 0 for offset.

I understand that, but I was wondering whether this was originally designed with the assumption that it would return everything.

Could forcing the limit to be applied affect the dashboard frontend or backend?

wilfredmulenga: Sorry for the late response. Looking at the design, pagination was taken into consideration so we might not face a scenario without pagination being explicitly defined for this case.

wilfredmulenga:
Closing this — the fixes here (bounded pagination, pipeline-form $lookup for topics filter) were absorbed into #2202 which rewrote the `findAllWithStarcountSort` and `findAllWithTopicsFilter` paths more thoroughly. No diff remains between this branch and main.

PR #2186, opened 2026-05-10
fix(server): add HTTP client timeouts for marketplace and plugin upload (REL-02, REL-03)

wilfredmulenga:
## Summary

- **REL-02**: Marketplace HTTP requests used `http.DefaultClient` (no timeout) and `http.NewRequest` (no context propagation). A slow or unresponsive marketplace would hang handler goroutines indefinitely.
- **REL-03**: Plugin upload from user-supplied URLs used `http.DefaultClient` with no transport-level hardening, leaving the server vulnerable to slow-drip/slowloris attacks.

## Changes

**`internal/infrastructure/marketplace/marketplace.go`**
- Added `marketplaceClient` with `Timeout: 30s` — replaces `http.DefaultClient`
- Fixed `client()` to pass the timeout client into the oauth2 context so the token fetch also has a deadline
- Switched `http.NewRequest` → `http.NewRequestWithContext` on both GET and POST calls

**`internal/usecase/interactor/plugin_upload.go`**
- Added `remotePluginClient` with a hardened `http.Transport`: 5s dial, 5s TLS handshake, 10s response header (kills slow-drip), 60s total timeout
- Redirect limit of 5
- Wrapped `res.Body` in `io.LimitReader` before passing to `PackageFromZip` to prevent memory spikes from oversized responses

## References

Reported in: https://github.com/eukarya-inc/compliance/issues/7 (REL-02, REL-03)


server/internal/usecase/interactor/plugin_upload.go
```diff
@@ -20,6 +23,28 @@ import (
...
+		DisableCompression:    true,
+	},
+	CheckRedirect: func(_ *http.Request, via []*http.Request) error {
+		if len(via) >= 5 {
+			return errors.New("too many redirects")
+		}
+		return nil
+	},
+}
+
```
soneda-yuya, line 51: Could you explain the rationale for why this configuration is appropriate?

server/internal/usecase/interactor/plugin_upload.go
```diff
@@ -20,6 +23,28 @@ import (
...
+// ResponseHeaderTimeout kills slow-drip connections that bypass the overall Timeout.
+// See: https://github.com/eukarya-inc/compliance/issues/7
+var remotePluginClient = &http.Client{
+	Timeout: 60 * time.Second,
+	Transport: &http.Transport{
+		DialContext:           (&net.Dialer{Timeout: 5 * time.Second, KeepAlive: 30 * time.Second}).DialContext,
+		TLSHandshakeTimeout:   5 * time.Second,
+		ResponseHeaderTimeout: 10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+		IdleConnTimeout:       90 * time.Second,
```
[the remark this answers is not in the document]
wilfredmulenga: fixed

server/internal/usecase/interactor/plugin_upload.go
wilfredmulenga: We limit to 5 redirects to avoid allowing the default 10, which could be unsafe if the plugin source is malicious.

server/internal/usecase/interactor/plugin_upload.go
```diff
@@ -20,6 +23,28 @@ import (
 
 var pluginPackageSizeLimit int64 = 10 * 1024 * 1024 // 10MB
 
+// remotePluginClient is hardened for fetching user-supplied URLs (REL-03).
+// ResponseHeaderTimeout kills slow-drip connections that bypass the overall Timeout.
```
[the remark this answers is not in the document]
wilfredmulenga: fixed

server/internal/usecase/interactor/plugin_upload.go
soneda-yuya: Thanks!
Sorry I don't understand why does system handle redirect?
When redirect happen on this client?
If redirects are not expected to happen, would setting this to 1 be better?

soneda-yuya: Also, I’m curious why each of the numbers here ended up being those values.
```
	Timeout: 60 * time.Second,
	Transport: &http.Transport{
		Proxy:                 http.ProxyFromEnvironment,
		DialContext:           (&net.Dialer{Timeout: 5 * time.Second, KeepAlive: 30 * time.Second}).DialContext,
		TLSHandshakeTimeout:   5 * time.Second,
		ResponseHeaderTimeout: 10 * time.Second,
		ExpectContinueTimeout: 1 * time.Second,
		IdleConnTimeout:       90 * time.Second,
		MaxIdleConns:          100,
		MaxIdleConnsPerHost:   10,
		DisableCompression:    true,
	},
```

wilfredmulenga: > Thanks! Sorry I don't understand why does system handle redirect? When redirect happen on this client? If redirects are not expected to happen, would setting this to 1 be better?

A limit of 1 might not be sufficient since the number of redirects can vary by plugin host. I settled on 5 as a balance between practicality and not being too permissive like the default 10.

wilfredmulenga: > Also, I’m curious why each of the numbers here ended up being those values.
> 
> ```
> 	Timeout: 60 * time.Second,
> 	Transport: &http.Transport{
> 		Proxy:                 http.ProxyFromEnvironment,
> 		DialContext:           (&net.Dialer{Timeout: 5 * time.Second, KeepAlive: 30 * time.Second}).DialContext,
> 		TLSHandshakeTimeout:   5 * time.Second,
> 		ResponseHeaderTimeout: 10 * time.Second,
> 		ExpectContinueTimeout: 1 * time.Second,
> 		IdleConnTimeout:       90 * time.Second,
> 		MaxIdleConns:          100,
> 		MaxIdleConnsPerHost:   10,
> 		DisableCompression:    true,
> 	},
> ```

Most of these are taken directly from Go's http.DefaultTransport values, https://go.dev/src/net/http/transport.go

I adjusted a few values like TLS, ResponseHeaderTimeout and Timeout to 60 to account for a plugin that might be relatively large file ( < 10MB ).

soneda-yuya: @wilfredmulenga 
What is the timeout for the Visualizer HTTP server?

At minimum, all of these timeout values should be shorter than that, because the HTTP server is the top-level layer of the communication flow.

Also, we should check the actual Visualizer performance using OTEL and set the timeouts based on the observed behavior

wilfredmulenga: Thanks @soneda-yuya -san, there is no explicit timeout on visualizer HTTP server. The Load balancer has 30s timeout so we could apply the same here https://github.com/eukarya-inc/infrastructure/blob/main/modules/reearth_visualizer_api/google_compute_backend_service.tf

Do you want us to add a timeout for visualizer?

I am yet to thoroughly integrate OTEL into my workflow. We could catch up about that one.

soneda-yuya: If the load balancer timeout is 30 seconds, shouldn’t the HTTP client timeout be shorter than or equal to that?

I think we should design these timeout values by considering the overall timeout chain.

Also, the Echo server has default timeout settings as well.

wilfredmulenga: Good question. Yes, we should set to to match the load balancer. I have updated the PR. 

The Echo server has zero timeout. Cloud run has 300s.

soneda-yuya (approved):
LGTM!

PR #2188, opened 2026-05-10
fix(server): deduplicate property schema lookups in plugin export (SCA-05)

wilfredmulenga:
Fixes N+1 property schema queries in `ExportPlugins`. Previously `FindByID` was called once per extension across all plugins. Replaced with a cache map so each unique schema is fetched at most once.

Tested locally — export of a project with 2 custom plugins (8 extensions) completes successfully.

See: https://github.com/eukarya-inc/compliance/issues/7 (SCA-05)


soneda-yuya (approved):
LGTM

PR #2254, opened 2026-06-08
fix(server): require auth on export download endpoint (SEC-02)

wilfredmulenga:
## Summary

- Add auth middleware to `/export/:filename` — previously unauthenticated, allowing anyone who knew the project ID to download private project exports
- Fix the delete-after-download goroutine to use `context.Background()` instead of the request context, which was already cancelled by the time the goroutine ran

### DIFF
--- a/server/internal/app/file.go
+++ b/server/internal/app/file.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/labstack/echo/v4"
+	"github.com/reearth/reearth/server/internal/adapter"
 	"github.com/reearth/reearth/server/internal/adapter/middleware"
 	"github.com/reearth/reearth/server/internal/usecase/gateway"
 	"github.com/reearth/reearth/server/pkg/id"
@@ -73,6 +74,10 @@ func serveFiles(
 	ec.GET(
 		"/export/:filename",
 		fileHandler(func(ctx echo.Context) (io.Reader, string, error) {
+			if adapter.Operator(ctx.Request().Context()) == nil {
+				return nil, "", echo.ErrUnauthorized
+			}
+
 			filename := ctx.Param("filename")
 
 			r, err := fileGateway.ReadExportProjectZip(ctx.Request().Context(), filename)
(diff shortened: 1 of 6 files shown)


server/internal/app/file.go
```diff
@@ -87,6 +94,7 @@ func serveFiles(
 			}()
 			return r, filename, nil
 		}),
+		exportAuthMiddleware,
 		middleware.FilesCORSMiddleware(domainChecker, allowedOrigins),
```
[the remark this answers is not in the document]
wilfredmulenga: fixed in 440e92694f8b437034cd692ff69c030c2750a4fa

server/internal/app/file.go
```diff
@@ -73,12 +82,10 @@ func serveFiles(
 			}
 			fmt.Printf("[export] download file: %s \n", filename)
 
-			rctx := ctx.Request().Context()
-
 			go func() {
 				// download and then delete
 				time.Sleep(3 * time.Second)
-				err := fileGateway.RemoveExportProjectZip(rctx, filename)
+				err := fileGateway.RemoveExportProjectZip(context.Background(), filename)
```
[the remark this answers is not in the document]
wilfredmulenga: fixed in 440e92694f8b437034cd692ff69c030c2750a4fa

server/internal/app/file.go
```diff
@@ -61,9 +64,20 @@ func serveFiles(
...
+		exportAuthMiddleware = attachOpMiddlewareReearthAccounts(cfg)
+	}
+
 	ec.GET(
 		"/export/:filename",
 		fileHandler(func(ctx echo.Context) (io.Reader, string, error) {
+			if adapter.Operator(ctx.Request().Context()) == nil {
+				return nil, "", echo.ErrUnauthorized
+			}
+
```
soneda-yuya, line 80: @wilfredmulenga  
2. Public projects will no longer be exportable. Is that okay?
3. How should we handle private projects in workspaces the user does not belong to?
I’m especially concerned about 1.

soneda-yuya (approved):

server/internal/app/file.go
wilfredmulenga: Thanks @soneda-yuya ,
1. That is on dashboard right? Will confirm but I think we would still need the possibility to export public projects. Thank you for spotting that.
2. Sorry I didn't fully understand this one. Do you mean a private project that they can view but belongs to a workspace they are not a member of? I am wondering how that one works. I will try reach out to confirm.

wilfredmulenga: Thanks. I confirmed that this change would indeed break the dashboard explore page download feature. I will mark it as a low risk as the project id is not guessable by brute force and close this change for now. Thank you for checking.

soneda-yuya: @wilfredmulenga 

>Sorry I didn't fully understand this one. Do you mean a private project that they can view but belongs to a workspace they are not a member of? I am wondering how that one works. I will try reach out to confirm.

Could you investigate how does this function work?

```golang
		if adapter.Operator(ctx.Request().Context()) == nil {
				return nil, "", echo.ErrUnauthorized
			}
```

PR #2285, opened 2026-06-30
refactor(web): improve UI/UX for tiles added by Street View and Google Search Widgets [VIZ-DEV-86]

mkumbobeaty:
# Overview
This PR removes tile configuration from the Street View widget and introduces automatic management of a required system tile for Google-based widgets (Google Search and Street View).

Instead of storing tile configuration inside the widget, the application now maintains a dedicated system tile within the Tiles panel. This tile is automatically created, updated, and removed based on the presence of Google widgets.

## What I've done
- Removed tile configuration from the Street View widget.
- Introduced a hidden category field for tile items to distinguish between user-managed (custom) and application-managed (system) tiles.
- Implemented automatic system tile management:
   - Creates a system tile when a Google Search or Street View widget is added and none exists.
   - Removes the system tile when no Google Search or Street View widgets remain.
- Updated the Tiles panel to clearly separate system-managed behavior from custom tiles:
   - The system tile is always kept at the end of the list.
   - New custom tiles are inserted before the system tile.
   - The system tile can only use `Google Satellite `or `Google Roadmap.`
   - `opacity` and `zoom level` settings are disabled for the system tile.
   - The system tile cannot be deleted manualy.
   - Drag-and-drop reordering is disabled while a system tile is present.

    ## What I haven't done

## Checklist


web/src/app/ui/fields/Properties/index.tsx
```diff
@@ -121,23 +142,32 @@ const PropertyItem: FC<Props> = ({
...
+   groups.length > 0 && isSystemTile(groups[groups.length - 1]);
+
   return (
     <FieldsWrapper>
       {isList && !!item && (
         <ListField
           title={item.title || (item.id === "default" ? "defaultItemName" : "")}
           items={propertyListItems}
           selected={selected}
+          dragDisabled={isDragDisabled}
```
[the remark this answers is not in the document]
mkumbobeaty: No the aim is when there. is system tile exist the entire tile list should be dissabled the drag amd drop of any item tile

web/src/app/features/Visualizer/utils/tilesMigration.ts
```diff
@@ -344,12 +236,9 @@ export function migrateViewerPropertyTiles(
...
 
-  // Check if Street View widget exists and needs tile appending (EE only)
-  const streetViewTiles = config.isEE ? extractStreetViewTiles(config.widgets) : [];
-  const needsStreetViewTile = streetViewTiles.length > 0;
 
   // Return original if nothing needs processing
-  if (!tilesNeedProcessing && !terrainNeedsProcessing && !terrainNeedsNormalMap && !needsStreetViewTile) {
+  if (!tilesNeedProcessing && !terrainNeedsProcessing && !terrainNeedsNormalMap) {
     return viewerProperty;
   }
```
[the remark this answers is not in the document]
mkumbobeaty: Thanks for pointing this out. We will handle this with a data migration. Existing scenes that have the widget/type stored but are missing the system tile will be migrated to add the required system tile. This avoids the need for a runtime fallback.

web/src/app/features/Editor/Map/InspectorPanel/hooks/usePropertyDecorations.tsx
```diff
@@ -138,6 +140,37 @@ export const usePropertyDecorations = () => {
         }
       }
 
+      // Business Rule: System tile type options restricted
+      // Only allow google_satellite and google_roadmap for system tiles
+      if (schemaId === "tile_type" && schemaGroup === "tiles") {
+        const tileCategoryField = internalFields?.find((f) => f.id === "tile_category");
+        if (tileCategoryField?.value === "system") {
```
ZTongci: Could we export SYSTEM_TILE_CATEGORY from useSystemTile.ts and import it in Properties/index.tsx and convert-object.ts instead of repeating the "system" string literal?

web/src/app/features/Editor/Widgets/WidgetManagerPanel/hooks.ts
```diff
@@ -27,32 +43,89 @@ export default ({ sceneId, selectWidget }: Props) => {
...
-      await addWidget(sceneId, id, type);
+      if (!sceneId || !id) return;
+
+      const shouldEnsureSystemTile =
+        isGSSVWidgetId(id) && !hasInstalledGSSVWidget();
+
+      await addWidget(sceneId, id, toWidgetAlignSystemType(widgetsViewDevice));
+
+      if (shouldEnsureSystemTile && !getSystemTileItemId()) {
+        await addSystemTile();
```
ZTongci, line 68: In handleWidgetAdd, the guard !hasInstalledGSSVWidget() reads from the stale installedWidgets cache before either mutation has esolved. If a user rapidly adds both Google Map Search and Street View (two in-flight calls), both will see no GSSV widget installed and both will set shouldEnsureSystemTile = true. The second check !getSystemTileItemId() also reads from a pre-mutation snapshot of scene?.property, so it won't catch the first call's tile either. Both calls end up invoking addSystemTile(), creating two system tiles in the Tiles panel.Probably rare in practice, but the guard doesn't actually prevent it. One option: move the system tile existence check inside addSystemTile itself using a fresh API fetch, rather than relying on the stale closure state.

ZTongci (commented):
PTAL! Thank you!

ZTongci (approved):

server/internal/infrastructure/mongo/migration/260701000000_set_tile_category.go
```diff
@@ -0,0 +1,229 @@
...
+							}
+						}
+						if streetViewLegacyTileTypes[tileType] && !hasTileCategory {
+							fmt.Printf("[migration] SetTileCategory: removing legacy tile tile_type=%q (property %q)\n",
+								tileType, doc.ID)
+							removedAny = true
+							continue
+						}
+						kept = append(kept, group)
+					}
```
airslice: I don't think we need to remove exist google map tiles, they were added by user by purpose, should be kept.

server/internal/infrastructure/mongo/migration/260701000000_set_tile_category.go
```diff
@@ -0,0 +1,229 @@
...
+						for _, field := range group.Fields {
+							if field.Field == "tile_category" {
+								if val, ok := field.Value.(string); ok && val == "system" {
+									hasSystemTile = true
+								}
+							}
+						}
+					}
+					if hasSystemTile {
+						fmt.Printf("[migration] SetTileCategory: property %q already has a system tile, skipping\n", doc.ID)
```
airslice, line 192: The street view widget might have the data for tiles (from ealier implement), we need to delete the property if it exist.

server/internal/infrastructure/mongo/migration/260701000000_set_tile_category.go
```diff
@@ -0,0 +1,229 @@
...
+						kept = append(kept, group)
+					}
+
+					// Add the canonical system tile.
+					newTile := &mongodoc.PropertyItemDocument{
+						Type:        "group",
+						ID:          id.NewPropertyItemID().String(),
+						SchemaGroup: "tiles",
+						Fields: []*mongodoc.PropertyFieldDocument{
+							{Field: "tile_type", Type: "string", Value: "google_satellite"},
```
airslice: I think we want to try get the tile type from street view widget's property (if exist)

server/internal/infrastructure/mongo/migration/260701000000_set_tile_category.go
```diff
@@ -0,0 +1,229 @@
...
+						Type:        "grouplist",
+						ID:          id.NewPropertyItemID().String(),
+						SchemaGroup: "tiles",
+						Groups: []*mongodoc.PropertyItemDocument{
+							{
+								Type:        "group",
+								ID:          id.NewPropertyItemID().String(),
+								SchemaGroup: "tiles",
+								Fields: []*mongodoc.PropertyFieldDocument{
+									{Field: "tile_type", Type: "string", Value: "google_satellite"},
```
airslice: Same as above, try get the tile type from street viw widget's property (if exist)

server/internal/infrastructure/mongo/migration/260701000000_set_tile_category.go
```diff
@@ -0,0 +1,229 @@
...
+									{Field: "tile_category", Type: "string", Value: "system"},
+								},
+							},
+						},
+					}
+					doc.Items = append(doc.Items, newTilesGroup)
+					fmt.Printf("[migration] SetTileCategory: created tiles group with system tile (property %q)\n", doc.ID)
+					modified = true
+				}
+
```
airslice, line 244: Also need to try clean up the widget property of tiles on street view widget (if exist)

server/internal/infrastructure/mongo/migration/260701000000_set_tile_category.go
```diff
@@ -0,0 +1,229 @@
+package migration
+
```
wilfredmulenga, line 2: Please add a revert migration file and tests for the migration.

server/internal/infrastructure/mongo/migration/260701000000_set_tile_category.go
```diff
@@ -0,0 +1,229 @@
...
+			for _, row := range rows {
+				var doc mongodoc.PropertyDocument
+				if err := bson.Unmarshal(row, &doc); err != nil {
+					var raw struct {
+						ID string `bson:"id"`
+					}
+					_ = bson.Unmarshal(row, &raw)
+					fmt.Printf("[migration] SetTileCategory: failed to unmarshal property id=%q: %v\n", raw.ID, err)
+					totalFailed++
+					return err
```
wilfredmulenga: mongox.Find propagates this error immediately and aborts the whole migration, not just this batch, so totalFailed can never exceed 1 in practice, and any already processed batches stay saved while later ones are silently never reached. Might be safer to log and skip this doc instead of returning err.

server/internal/infrastructure/mongo/migration/260701000000_set_tile_category.go
```diff
@@ -0,0 +1,229 @@
...
+					ids = append(ids, doc.ID)
+					newRows = append(newRows, doc)
+					totalMigrated++
+				} else {
+					totalSkipped++
+				}
+			}
+
+			if len(ids) > 0 {
+				return propCol.SaveAll(ctx, ids, newRows)
```
wilfredmulenga, line 257: Consider logging ids before this SaveAll call, same as the revert migration for the Cesium Ion tile change does. That gives a record of exactly which properties were about to be touched in this batch before the write happens, which helps a lot if a later batch fails partway through.

web/src/app/ui/fields/Properties/index.tsx
```diff
@@ -143,6 +150,8 @@ const PropertyItem: FC<Props> = ({
     [item?.schemaFields, item?.schemaGroup, selectedItem]
   );
 
+  console.log("PropertyItem selectedItem:", selectedItem);
```
wilfredmulenga: We can remove this console log

server/internal/infrastructure/mongo/migration/260701000000_set_tile_category.go
```diff
@@ -0,0 +1,333 @@
...
+
+	// Step 3: Remove legacy "tiles" grouplist from streetView/googleMapSearch widget properties.
+	// An earlier implementation stored tile configuration directly on the widget property.
+	// That data is now managed at the scene level, so we clean up the stale widget property groups.
+	if len(widgetPropertyIDs) == 0 {
+		fmt.Println("[migration] SetTileCategory: no widget properties to clean up")
+		return nil
+	}
+
+	fmt.Printf("[migration] SetTileCategory: cleaning up legacy tiles group from %d widget properties\n", len(widgetPropertyIDs))
```
wilfredmulenga, line 277: I would consider something like this to log to recover the original `tile_type` as the revert migration does cover this change

`fmt.Printf("[migration] SetTileCategory: widget property %q had tile_type=%q before removal\n", doc.ID, val)`

airslice (approved):

PR #2293, opened 2026-07-07
refactor(web): localize preset style categories and styles[VIZ-DEV-94]

mkumbobeaty:
# Overview
This Pr supports translations for preset style categories and styles
## Checklist


airslice:
Please update handleLayerStyleAddition as well, when app run in Ja, it should create new layer style with Ja names. Please check related logic as well, for example selection, please do refactor when necessary.

web/src/app/features/Editor/Map/LayerStylePanel/PresetLayerStyle/index.tsx
```diff
@@ -86,7 +86,7 @@ const PresetLayerStyle: FC<PresetLayerStyleProps> = ({
       if (config.type === "empty") {
         return {
           id: "empty",
-          title: "Empty",
+          title: lang === "ja" ? "空" : "Empty",
           onClick: () => handleLayerStyleAddition({}),
           dataTestid: "preset-style-empty"
```
ZTongci, line 98: When `lang === "ja"`, the "Empty" item displays as "空" but `handleLayerStyleAddition` is called without a `styleName`, so the created style is saved as "Style 1" instead of "空". Other presets pass the title as the second argument — same should apply here.

```ts
onClick: () => handleLayerStyleAddition({}, lang === "ja" ? "空" : "Empty")
```

web/src/app/features/Editor/Map/LayerStylePanel/PresetLayerStyle/index.tsx
```diff
@@ -111,7 +111,7 @@ const PresetLayerStyle: FC<PresetLayerStyleProps> = ({
       const category = config.preset;
       return {
         id: category.id,
-        title: category.title,
+        title: lang === "ja" && category.titleJa ? category.titleJa : category.title,
         icon: "folderSimple",
         dataTestid: category.testId,
```
ZTongci, line 119: `lang === "ja" && x.titleJa ? x.titleJa : x.title` is repeated 4 times. Consider a small helper:

```ts
const localize = (title: string, titleJa?: string) =>
  lang === "ja" && titleJa ? titleJa : title;
```

ZTongci (approved):

.github/workflows/playwright_ui_tests.yml
```diff
@@ -62,6 +62,7 @@ jobs:
         run: npm ci
 
       - name: Run API setup (writes .auth/api-token.json with dev token)
+        continue-on-error: true
```
wilfredmulenga: This change is not required

airslice (approved):

PR #2304, opened 2026-07-15
ci: added video recording for the visualizer tests

waqaszulfiqar:
# Overview

## What I've done
Pass recordVideo on the non-IAP path and centralize the options in a shared RECORD_VIDEO_OPTIONS constant so every context-creation path records video consistently.

## Checklist


wilfredmulenga:
Thanks for adding this in. I am wondering if the videos get pushed to the Allure report and whether we would like to push on videos ( for both passed and failed tests ) to allure.

airslice (approved):

PR #2332, opened 2026-07-30
fix(web): plugin playground import: filter out macOS metadata and handle nested file properly [VIZ-DEV-153]

airslice:
# Overview

Fix plugin zip import in Plugin Playground to handle edge cases:

  - Filter out macOS `__MACOSX/` folder and `._` prefixed metadata files, also .DS_Store
  - Limit import to root-level files or files within a single top-level folder
  - Extract filename from last path segment instead of assuming index [1]

  Previously, importing a zip created on macOS would include hidden metadata
  files (e.g., `._map-quiz.js`), doubling the file count. Additionally, zips
  without a containing folder would cause "Cannot read properties of undefined"
  errors when accessing `split("/")[1]` on root-level filenames.

## Checklist


wilfredmulenga (approved):
Looks good

PR #2334, opened 2026-07-30
fix(web): improve navigation UI/UX across the application

mkumbobeaty:
# Overview
This PR refactors the Visualizer navigation to align with the Re:Earth ecosystem, providing a more consistent and intuitive experience for navigating between projects, workspaces, Re:Earth products, and account-related actions across the application.

## What I've done
* Refactored the Dashboard and Navbar navigation to follow the common Re:Earth navigation experience.
* Unified the profile and workspace navigation.
* Improved project and workspace switching.
* Added consistent navigation between Re:Earth products.
* Improved access to account settings, workspace management, documentation, and logout.

## Checklist


web/src/app/features/Dashboard/LeftSidePanel/index.tsx
```diff
@@ -38,15 +46,22 @@ const LeftSidePanel: FC<Props> = ({
   return (
     <SidebarWrapper data-testid="left-side-panel">
       <SidebarMainSection data-testid="sidebar-main-section">
-        <LogoWrapper data-testid="sidebar-logo" />
+        <SidebarTopSection data-testid="sidebar-top-section">
+          <LogoWrapper data-testid="sidebar-logo" />
+          <ProductsMenu workspaceId={currentWorkspace?.id} />
```
airslice: We don't need this when `REEARTH_WEB_FEATURE_COLLECTION` is not `ee`

web/src/app/features/ProductsMenu/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+      background: "#4B3F22",
+      onNavigate: cmsUrl ? () => openUrlInNewTab(cmsUrl) : undefined
+    },
+    {
+      id: "flow",
+      title: t("Flow"),
+      icon: "flowLogo",
+      background: "#2C3F34",
+      disabled: true
+    }
```
airslice: Please remove this for now.

web/src/app/features/ProductsMenu/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+  ];
+
+  const otherLinks: MenuItem[] = [
+    {
+      id: "home",
+      title: t("Re:Earth Home"),
+      icon: "home",
+      onNavigate: goToDashboard
```
airslice: Re:Earth home is not dashboard page, please add a new env var for it.

web/src/services/config/index.ts
```diff
@@ -87,6 +87,7 @@ export type Config = {
   platformUrl?: string;
   tileServerBaseUrl?: string;
   tileServerToken?: string;
+  cmsUrl?: string;
```
airslice, line 90: For any new env var please update .env.example for it.

web/src/app/features/Navbar/LeftSection/index.tsx
```diff
@@ -80,23 +136,33 @@ const LeftSection: React.FC<Props> = ({
...
+          <Icon
+            icon="caretDown"
+            size="small"
+            data-testid="profile-caretDownIcon"
+          />
+        }
+        menu={accountMenuItems}
+        dataTestid="avatar-popupMenu"
+      />
+      <ProductsMenu workspaceId={currentWorkspace?.id} />
```
airslice: We don't need ProductsMenu when !isEE

web/src/app/features/Dashboard/LeftSidePanel/Avatar.tsx
```diff
@@ -0,0 +1,60 @@
...
+            <AvatarImage
+              src={avatarURL}
+              alt="Avatar"
+              onError={() => setShowAvatar(false)}
+            />
+          ) : (
+            <Typography size="body" data-testid="profile-avatar-initial">
+              {userName?.charAt(0).toUpperCase()}
+            </Typography>
+          )}
```
wilfredmulenga, line 30: `useState(!!avatarURL)` derives state from a prop, so it is only correct as long as the prop never changes. Here it does change without a remount: switching workspace calls `navigate("/dashboard/${workspaceId}")` (`Dashboard/hooks.ts`), which keeps the same route and element tree, so `AvatarWrapper` stays mounted while `avatarURL` flips between `metadata.photoURL` and `workspace.photoURL`.

So if the workspace you land on has no avatar, `showAvatar` initializes to `false` and nothing ever sets it back to `true`. Switch to a workspace that does have one and the `avatarURL && showAvatar` guard is still false, so the image never renders and you keep seeing the initial letter. Same thing after a failed load: once `onError` fires, a later valid URL stays suppressed.

(Mount is fine, to be clear. `Page` gates on `loadingMe`, so `avatarURL` is already resolved the first time this renders.)

Tracking which URL failed instead of a bare boolean fixes both cases, and the flag then resets on its own whenever the URL changes:

```suggestion
  const [failedURL, setFailedURL] = useState<string>();
  const showImage = !!avatarURL && avatarURL !== failedURL;
  const popupMenu = useAvatarMenuItems({ userName, userEmail, onSignOut });

  return (
    <PopupMenu
      label={
        <Avatar data-testid="profile-avatar">
          {showImage ? (
            <AvatarImage
              src={avatarURL}
              alt="Avatar"
              onError={() => setFailedURL(avatarURL)}
            />
          ) : (
            <Typography size="body" data-testid="profile-avatar-initial">
              {userName?.charAt(0).toUpperCase()}
            </Typography>
          )}
```

web/src/app/features/Dashboard/LeftSidePanel/Avatar.tsx
```diff
@@ -0,0 +1,60 @@
...
+          {avatarURL && showAvatar ? (
+            <AvatarImage
+              src={avatarURL}
+              alt="Avatar"
+              onError={() => setShowAvatar(false)}
+            />
+          ) : (
+            <Typography size="body" data-testid="profile-avatar-initial">
+              {userName?.charAt(0).toUpperCase()}
+            </Typography>
```
wilfredmulenga, line 29: Minor, non-blocking. `userName` cannot actually be undefined here: `Page` waits on `loadingMe` before rendering Dashboard, and `name` is `String!` in the schema, so the `?.` is defensive rather than load-bearing. The gap that is still reachable is an empty or whitespace-only name, where `charAt(0)` returns `""` and the circle renders blank with no letter and no image.

Naming the value handles that in one place and makes the intent clearer than an optional chain that quietly resolves to nothing:

```suggestion
            <Typography size="body" data-testid="profile-avatar-initial">
              {initial}
            </Typography>
```

with `const initial = userName?.trim().charAt(0).toUpperCase() || "?";` next to the other derived values above the return. The `trim()` only matters if display names can come back padded, so drop it if we already normalize on the way in.

web/src/app/features/Dashboard/LeftSidePanel/Avatar.tsx
```diff
@@ -0,0 +1,60 @@
...
+          )}
+        </Avatar>
+      }
+      menu={popupMenu}
+      dataTestid="avatar-popupMenu"
```
wilfredmulenga, line 34: Nit: test ids in the app are almost all kebab-case (13 camelCase out of ~540). This PR adds several of them: `avatar-popupMenu` here and in `Navbar/LeftSection`, plus `avatar-userInfo`, `avatar-accountSettings`, `avatar-signOut`, `profile-newProject` and `profile-newWorkspace`. Nothing references any of them yet, so they are free to rename.

```suggestion
      dataTestid="avatar-popup-menu"
```

PR #2348, opened 2026-08-03
fix(server): apply workspace/scene filters to authenticated requests (SEC-01)

wilfredmulenga:
## Summary
UsecaseMiddleware ran globally, before the auth middleware, so its operator was always nil and the workspace/scene filters never got applied to the repos. Any authenticated user could read another workspace's projects and scenes, private ones included, since the repos ended up unrestricted.

The fix re-runs the same middleware on the private route group, after the operator is attached, so authenticated requests get their usecase container rebuilt with the operator's filters actually applied. Public routes keep the original global registration, unchanged.

## Test plan
- [x] Unit test on UsecaseMiddleware pinning the filtering behavior with and without an operator on the context
- [x] End to end test through the real HTTP server, confirmed to fail against the unpatched code (returned another workspace's project) and pass with the fix
- [x] Confirmed the workspace owner can still read their own projects after the fix


airslice (approved):
Non-blocking: this fixes the immediate issue, but having UsecaseMiddleware registered globally and then re-run for authenticated routes makes the authorization behavior somewhat dependent on middleware ordering.
It might be worth considering later whether public/private usecase construction can be separated more explicitly, so authenticated routes never receive an unrestricted container in the first place.

wilfredmulenga:
> Non-blocking: this fixes the immediate issue, but having UsecaseMiddleware registered globally and then re-run for authenticated routes makes the authorization behavior somewhat dependent on middleware ordering. It might be worth considering later whether public/private usecase construction can be separated more explicitly, so authenticated routes never receive an unrestricted container in the first place.

Thanks. I will create a task to refactor this.

PR #2358, opened 2026-08-05
fix(server): restore asset URL prefix check in project export

wilfredmulenga:
## Summary
`IsCurrentHostAssets` required a URL to start with both `assets/` and the current host at the same time, which no real absolute URL can satisfy. In any environment with a host configured, this made `AddZipAsset` treat every real asset reference as not being an asset at all, so project export silently included no assets.

This was introduced by #1801 ("Added GCS Read permission error"), which changed the original `||`-based check to `&&` and dropped its fallback line. This restores the original logic: a relative `assets/` or `/assets` path, or an absolute URL under the current host, is recognized as an asset reference.

## Test plan
- [x] `go build ./...` passes
- [x] New regression test `TestIsCurrentHostAssets` covers relative paths, rooted paths, and absolute URLs under the current host (and correctly rejects a different host)
- [x] Verified the test fails when the bug is reintroduced, confirming it catches the regression
- [x] Confirmed `AddZipAsset` is the only caller of this function, so the fix is scoped to project export

### DIFF
--- a/server/internal/usecase/interactor/common.go
+++ b/server/internal/usecase/interactor/common.go
@@ -301,10 +301,10 @@ func (d ProjectDeleter) Delete(ctx context.Context, prj *project.Project, force 
 }
 
 func IsCurrentHostAssets(ctx context.Context, u string) bool {
-	if strings.HasPrefix(u, "assets/") && strings.HasPrefix(u, adapter.CurrentHost(ctx)) {
+	if strings.HasPrefix(u, "assets/") || strings.HasPrefix(u, "/assets") {
 		return true
 	}
-	return false
+	return strings.HasPrefix(u, adapter.CurrentHost(ctx))
 }
 
 func ReplaceToCurrentHost(ctx context.Context, urlString string) string {
--- a/server/internal/usecase/interactor/common.go
+++ b/server/internal/usecase/interactor/common.go
@@ -331,7 +331,9 @@ func IsCurrentHostAssets(ctx context.Context, u string) bool {
 	if err != nil {
 		return false
 	}
-	return parsedURL.Scheme == parsedHost.Scheme && parsedURL.Host == parsedHost.Host
+	return parsedURL.Scheme == parsedHost.Scheme &&
+		parsedURL.Host == parsedHost.Host &&
+		strings.HasPrefix(parsedURL.Path, "/assets/")
 }
 
 func ReplaceToCurrentHost(ctx context.Context, urlString string) string {
--- a/server/internal/usecase/interactor/common.go
+++ b/server/internal/usecase/interactor/common.go
@@ -259,6 +259,19 @@ type ProjectDeleter struct {
 	Asset           repo.Asset
 }
 
+// Delete runs the storage deletes (assets, plugin files, built scene, via
+// gateway.File -- GCS, S3, or local depending on deployment) inside the same
+// DB transaction as the Mongo removes. This means a transaction abort after
+// a storage delete already ran (e.g. a later step failing, or the
+// transaction hitting its time limit) can restore the project/scene rows
+// while the storage they point to is already gone (compliance scan REL-01).
+// We're accepting this tradeoff for now: splitting storage cleanup out of the
+// transaction would mean a delete that fails partway through can leave a
+// project's DB rows removed but some of its assets still in storage --
+// orphaned with no owning record. Given how rarely project deletes fail
+// partway through in practice, keeping the database-side deletion
+// transactional (at the cost of the rarer storage-outlives-rollback case
+// above) is the safer default.
 func (d ProjectDeleter) Delete(ctx context.Context, prj *project.Project, force bool, operator *usecase.Operator) error {
 	if prj == nil {
 		return nil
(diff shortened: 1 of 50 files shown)


server/internal/usecase/interactor/common.go
```diff
@@ -301,10 +301,10 @@ func (d ProjectDeleter) Delete(ctx context.Context, prj *project.Project, force
 }
 
 func IsCurrentHostAssets(ctx context.Context, u string) bool {
-	if strings.HasPrefix(u, "assets/") && strings.HasPrefix(u, adapter.CurrentHost(ctx)) {
+	if strings.HasPrefix(u, "assets/") || strings.HasPrefix(u, "/assets") {
 		return true
 	}
-	return false
+	return strings.HasPrefix(u, adapter.CurrentHost(ctx))
 }
```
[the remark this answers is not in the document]
wilfredmulenga: Fixed in 65b810dbe: guarded against an empty current host (returns false rather than matching everything), and replaced the raw string-prefix comparison with parsing both URLs and comparing scheme+host directly, so a lookalike domain like `api.example.com.evil.com` no longer matches. Added `TestIsCurrentHostAssets_NoCurrentHost` and a lookalike-domain case in `TestIsCurrentHostAssets`, both verified to fail with the old prefix-matching logic.

server/internal/usecase/interactor/common_test.go
```diff
@@ -0,0 +1,23 @@
...
+// satisfy -- so exports and asset replacement silently treated every real asset URL as not being
+// an asset at all whenever a host was actually configured. This confirms relative asset paths and
+// absolute URLs under the current host are both recognized.
+func TestIsCurrentHostAssets(t *testing.T) {
+	ctx := adapter.AttachCurrentHost(context.Background(), "https://api.example.com")
+
+	assert.True(t, IsCurrentHostAssets(ctx, "assets/foo.png"), "a relative assets/ path must be recognized")
+	assert.True(t, IsCurrentHostAssets(ctx, "/assets/foo.png"), "a rooted /assets path must be recognized")
+	assert.True(t, IsCurrentHostAssets(ctx, "https://api.example.com/assets/foo.png"), "an absolute URL under the current host must be recognized")
+	assert.False(t, IsCurrentHostAssets(ctx, "https://other-host.com/assets/foo.png"), "an absolute URL under a different host must not be recognized")
```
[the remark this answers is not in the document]
wilfredmulenga: Added in 65b810dbe: `TestIsCurrentHostAssets_NoCurrentHost` covers the empty-host case, and a lookalike-domain case (`api.example.com.evil.com`) was added to `TestIsCurrentHostAssets`. Both fail against the old raw-prefix-matching implementation.

airslice (commented):
I think the relative-path fix currently only works at the helper level, but not through the actual export flow.
IsCurrentHostAssets now correctly accepts assets/... and /assets/.... However, SearchAssetURL only calls AddZipAsset when cleanedStr starts with adapter.CurrentHost(ctx). With a configured host, relative paths fail this check and therefore never reach IsCurrentHostAssets.
For example, with CurrentHost = "https://example.com", assets/foo.png is accepted by IsCurrentHostAssets, but skipped by SearchAssetURL.
Could we avoid duplicating the filtering logic here and let IsCurrentHostAssets / AddZipAsset be the single source of truth? It would also be good to add a regression test through SearchAssetURL or SaveExportProjectZip to verify that a relative asset is actually included in the exported ZIP.

please let me know if i misunderstood something.
BTW this PR modify CHANGELOG.md directly, is this intential?

wilfredmulenga:
Good catch, you're right. Pushed a fix: dropped SearchAssetURL's own `strings.HasPrefix(cleanedStr, currentHost)` check and let `AddZipAsset`/`IsCurrentHostAssets` be the single source of truth for recognizing an asset reference. Added a regression test that goes through `SearchAssetURL` with a configured host so it actually exercises the path you flagged (confirmed it fails without the fix, passes with it).

The CHANGELOG.md diff was unintentional drift from an earlier merge of main into this branch, not something this PR meant to touch -- it's gone now.

ZTongci (approved):

server/internal/usecase/interactor/common.go
```diff
@@ -314,10 +314,24 @@ func (d ProjectDeleter) Delete(ctx context.Context, prj *project.Project, force
...
 
 func IsCurrentHostAssets(ctx context.Context, u string) bool {
-	if strings.HasPrefix(u, "assets/") && strings.HasPrefix(u, adapter.CurrentHost(ctx)) {
+	if strings.HasPrefix(u, "assets/") || strings.HasPrefix(u, "/assets") {
```
airslice:
```suggestion
	if strings.HasPrefix(u, "assets/") || strings.HasPrefix(u, "/assets/") {
```
Will it be better have this `/` in the end? To avoid something like `/assets-foo.png` passing this check?

wilfredmulenga: Good catch, fixed. Added `/assets-foo.png` as a regression case in TestIsCurrentHostAssets confirming it's rejected (verified it fails without this fix). Real asset URLs are always `assets/<file>` or `/assets/<file>` (the route is registered at `/assets/:filename`), so requiring the trailing slash doesn't reject anything legitimate.

wilfredmulenga: fixed

airslice (approved):

PR #2361, opened 2026-08-05
fix(server): batch layer property loads in the scene builder

wilfredmulenga:
## Problem

The scene builder loaded each layer's infobox, photo overlay and infobox block properties one ID at a time while building that layer's JSON. A scene of L layers holding B infobox blocks each cost roughly L*(2+B) queries, so a scene with hundreds of layers issued hundreds of round trips on every read (SCA-02).

## Fix

Collect every property ID from the layer list in one walk and load them before any JSON is built, then pass the result down. This is what scene- and story-level properties already do.

- Loads go out in fixed-size chunks, since the layer count is unbounded and a single call would trade many small queries for one arbitrarily large query and result set.
- IDs are deduplicated, so a property shared by several layers is fetched once.
- The loaded properties are keyed by ID rather than scanned linearly. A large scene has thousands of them, and scanning per lookup would have replaced the query cost with a quadratic one.

Load failures stay tolerated rather than returned, matching the previous behaviour where the per-layer loader error was discarded. A property that cannot be loaded is absent and renders empty instead of failing the whole scene build. Changing that would alter how publish behaves on partial failure, which is a separate decision.

## Verification

- go build, go vet, gofmt and go test ./... all pass (46 packages, no failures)
- New tests use a counting loader to assert on query count: batching to a single load, chunking a set larger than the chunk size without dropping or repeating IDs, deduplicating a shared property, tolerating a load failure, and issuing no query when a scene has no layers

## Not included

No limit on layers per scene. That is a product decision and not a substitute for batching, since a capped scene still issues one load per layer.

---

VIZ-DEV-164


server/pkg/scene/builder/nlsLayer.go
```diff
@@ -85,6 +85,14 @@ type geometryCollectionJSON struct {
...
 }
 
+// nlsLayersJSON builds each layer's JSON one at a time, and getNLSLayerJSON below loads that
+// layer's infobox/photo overlay/infobox block properties via b.ploader with a single ID per call
+// (SCA-02, compliance scan issue #96), instead of collecting every layer's property IDs up front
+// and issuing one batched b.ploader(ctx, ids...) call the way scene- and story-level properties
+// already do (see builder.go). For a scene with hundreds of layers this means hundreds of
+// individual property loads instead of one. Left as a comment for now rather than fixed: this is
+// a performance concern, not a security one, and there's currently no enforced limit on the
+// number of layers a scene can have, so batch this loading if that assumption changes.
```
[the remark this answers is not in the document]
wilfredmulenga: updated the comment

server/pkg/scene/builder/nlsLayer.go
```diff
@@ -85,6 +85,38 @@ type geometryCollectionJSON struct {
...
+// fixed-size chunks rather than a single call: the layer count is unbounded, so one batched load
+// would trade many small queries for one arbitrarily large query and result set. Chunking bounds
+// query size and peak memory without limiting what a scene can contain.
+//
+// 2. Tree traversal. b.nlsloader is called once per layer group to fetch that group's children,
+// and a group's children are not known until the group itself has been loaded, so this is a real
+// discover-then-fetch cycle that batching properties does not touch. It can still be reduced by
+// walking the tree level by level, collecting every group at one depth and issuing a single
+// nlsloader call for all of their children, which makes the traversal cost track tree depth
+// rather than group count.
```
airslice: The scene’s complete NLS layer list has already been loaded by FindByScene and passed through WithNLSLayers, while each group already contains its child IDs. Therefore this is not an inherent discover-then-fetch cycle. Future optimization should reuse/index the preloaded list and eliminate recursive repository loads, rather than issue one query per tree depth.

wilfredmulenga: You're right, thanks. Verified: `lg.Children().Layers()` already holds the child IDs, and the children themselves are already in memory too, since `NLSLayer.FindByScene` has no root predicate and every caller passes the full list to `WithNLSLayers`. `NLSLayerLoaderFrom` is an uncached `FindByIDs`, so those calls were refetching rows we held. No discover-then-fetch cycle, and batching per depth was the wrong target.

It turned out to be moot as well: nothing can create a layer group today, so that branch issues no queries at all. `ImportNLSLayers` forces `.Simple()` and discards children as unsupported, no mutation constructs one, and `gql/newlayer.graphql` still says `TODO: Make LayerGroup Real`.

So I replaced that section rather than reworded it. It now notes the traversal is queryless today and, if groups are ever made real, points at indexing `*b.nlsLayer` on ID as you suggested, plus a root filter for the top-level loop.

The PR has also grown into the actual fix for the one live N+1: layer property IDs are now collected in one walk and loaded in chunks, keyed by ID rather than scanned. Worth a re-read as a fix rather than a comment.

server/pkg/scene/builder/nlsLayer.go
```diff
@@ -85,23 +87,101 @@ type geometryCollectionJSON struct {
...
+		return nil
+	}
+
+	res := make(map[id.PropertyID]*property.Property, len(ids))
+	for start := 0; start < len(ids); start += propertyLoadChunkSize {
+		end := min(start+propertyLoadChunkSize, len(ids))
+		loaded, err := b.ploader(ctx, ids[start:end]...)
+		if err != nil {
+			continue
+		}
```
airslice, line 173: The previous per-property implementation ignored the loader error but still used any property returned alongside it. With chunking, a loader response containing partial results plus an error now causes the whole chunk to be discarded, so valid properties in the same chunk render empty. Could we merge non-empty results before handling the error, or fall back to per-ID loads for a failed chunk, and add a regression test for this partial-result-plus-error case?

airslice (approved):

server/pkg/scene/builder/nlsLayer.go
wilfredmulenga: Good catch. Fixed in 1eabec383: the loader now keeps any partial results returned alongside an error, then retries the still-missing IDs one at a time, so a single unloadable property no longer blanks the other valid ones in its chunk. Since the real property loader returns all-or-nothing on error, the per-ID fallback is what restores the pre-batch resilience. Added a regression test (TestLoadNLSLayerProperties_ChunkFailureFallsBackPerID) for the failed-chunk case.

airslice (approved):

PR #2363, opened 2026-08-05
fix(server): drop unused domains.domain branch from public project lookup (SCA-03)

wilfredmulenga:
## Summary
FindByPublicName matched on either alias or domains.domain. Checked the whole codebase (schema, mongo docs, every write path) and confirmed no project document has ever had a domains field. That branch of the $or could never match anything, but MongoDB still had to fall back to a collection scan for it on every call, since there's no index on that field.

This is a hot path: roughly 77k requests/day to /api/published/:name in prod (checked Cloud Run request logs), so this collection scan runs constantly.

## Fix
Dropped the unused domains.domain branch. The remaining filter (alias + publishmentstatus) is fully covered by the existing alias,publishmentstatus compound index, so the query now uses an index scan instead of a collection scan.

## Test plan
- [x] `go build ./...` passes
- [x] Existing `TestProject_FindByPublicName` still passes
- [x] Added `TestProject_FindByPublicName_UsesIndex`, which runs `explain` on the exact filter the method builds and asserts the winning plan isn't COLLSCAN
- [x] Confirmed the same filter shape produces a COLLSCAN before this fix
- [x] Confirmed the only callers (Published.Metadata and Published.Index) are unaffected in behavior, only performance

### DIFF
--- a/server/internal/infrastructure/mongo/project_test.go
+++ b/server/internal/infrastructure/mongo/project_test.go
@@ -214,6 +214,7 @@ func TestProject_FindByPublicName_UsesIndex(t *testing.T) {
 	}
 
 	assert.False(t, hasStage(winningPlan, "COLLSCAN"), "expected an index scan (no COLLSCAN stage), got: %v", winningPlan)
+}
 
 func TestProject_FindStarredByWorkspace(t *testing.T) {
 	c := mongotest.Connect(t)(t)
--- a/server/internal/infrastructure/mongo/project_test.go
+++ b/server/internal/infrastructure/mongo/project_test.go
@@ -176,8 +176,44 @@ func TestProject_FindByPublicName_UsesIndex(t *testing.T) {
 	winningPlan, ok := queryPlanner["winningPlan"].(bson.M)
 	require.True(t, ok, "queryPlanner missing winningPlan: %v", queryPlanner)
 
-	assert.NotEqual(t, "COLLSCAN", winningPlan["stage"], "expected an index scan, got a full collection scan: %v", winningPlan)
-}
+	var hasStage func(v any, stage string) bool
+	hasStage = func(v any, stage string) bool {
+		switch x := v.(type) {
+		case bson.M:
+			if s, ok := x["stage"].(string); ok && s == stage {
+				return true
+			}
+			for _, vv := range x {
+				if hasStage(vv, stage) {
+					return true
+				}
+			}
+		case map[string]any:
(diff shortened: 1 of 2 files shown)


airslice (approved):
LGTM!
BTW could you help check where does `domains.domain` comes from?

server/internal/infrastructure/mongo/project_test.go
```diff
@@ -143,6 +143,79 @@ func TestProject_FindByPublicName(t *testing.T) {
...
+	c := mongotest.Connect(t)(t)
+	ctx := context.Background()
+	client := mongox.NewClientWithDatabase(c)
+
+	r := NewProject(client)
+	require.NoError(t, r.Init(ctx))
+
+	prj := project.New().NewID().Workspace(accountsID.NewWorkspaceID()).Alias("explain-alias").PublishmentStatus(project.PublishmentStatusPublic).MustBuild()
+	_, err := c.Collection("project").InsertOne(ctx, util.DR(mongodoc.NewProject(prj)))
+	require.NoError(t, err)
```
ZTongci, line 173: This test only inserts one document. When the collection is tiny, MongoDB's query planner may choose COLLSCAN anyway because scanning one document is cheaper than using an index. The assert.False(t, hasStage(winningPlan, "COLLSCAN")) could pass for the wrong reason or fail on different MongoDB versions. Inserting more documents (e.g. 100+) before running explain would force the optimizer to prefer the index.

ZTongci (approved):

mkumbobeaty (approved):

server/internal/infrastructure/mongo/project_test.go
wilfredmulenga: Checked this directly against a real Mongo 8.0 instance with the test exactly as written (1 document): the planner already picked FETCH/IXSCAN, not COLLSCAN, so this specific failure mode didn't reproduce here -- with only one index matching the query's equality predicates, Mongo uses it directly rather than running its cost-based multi-plan trial (that trial is what would make a tiny collection favor a scan).

That said, the underlying caution is fair -- asserting on explain() output is inherently more sensitive to planner heuristics and Mongo version than asserting on observable results, so "doesn't fail today, on this version" isn't the same as "never will." Added 100 decoy documents before the explain call as a free hedge regardless; confirmed it adds no meaningful time to the test (~0.14s either way).

airslice (approved):

PR #2365, opened 2026-08-05
fix(e2e): restore two skipped UI tests and replace fixme plugin test

ZTongci:
# Overview

Restore two previously skipped Playwright UI tests in `projects.spec.ts` and replace a `test.fixme` placeholder in the plugin API spec with a proper assertion.

## What I've done

- **Unskipped** `"Should verify sketch tools are visible after adding style"` in `tests/project/projects.spec.ts`. This test is purely DOM-based (checks toolbar button visibility) with no canvas interaction, so it is safe to run in headless WebKit.

- **Unskipped** `"Should navigate to Scene panel and verify scene items"` in `tests/project/projects.spec.ts`.
  - Fixed `getSceneItemByName()` locator in `pages/projectScreenPage.ts` — changed from `[data-testid="editor-map-scene-item"] div:has-text("...")` (matched 2 descendant divs) to `[data-testid="editor-map-scene-item"]:has-text("...")` (matches exactly the outer wrapper).
  - Removed the erroneous `scenePanel.click()` call. The Scene panel renders inside a `Collapse` component that is **open by default** (fresh localStorage). Clicking its header was toggling it **closed**, hiding all items and causing the assertion to time out with zero matches.

- **Replaced `test.fixme`** in `api/tests/plugin.api.spec.ts`: the old placeholder attempted to uninstall the built-in `reearth` plugin, which the server intentionally rejects. The new test asserts that this rejection is correctly enforced (i.e., `gqlClient.mutate(UNINSTALL_PLUGIN, { pluginId: "reearth" })` must throw).

## What I haven't done

- Canvas-dependent drawing tests (`Should draw a polyline/polygon/circle/square`) remain skipped — these use `page.locator("canvas").click()` at pixel coordinates, which is unreliable in headless WebKit.
- `accountWorkspaceSettings.spec.ts` and `members.spec.ts` remain fully skipped (EE config / feature flag dependencies).
- The plugin API test cannot be verified locally due to Auth0 audience mismatch in the local dev environment; it is logically correct and verified structurally.

## How I tested

```bash
npx playwright test --project=webkit tests/project/projects.spec.ts
```

Result: **9 passed, 4 skipped, 0 failed**. Both unskipped tests pass. The 4 remaining skips are the canvas drawing tests.

- The removal of `scenePanel.click()` — confirm the Scene panel is always rendered open by default in the editor layout and does not require an explicit navigation action.
- The `plugin.api.spec.ts` negative assertion — confirm that the server's rejection of built-in plugin uninstall is the correct expected behavior to assert.

## Memo

These tests were originally skipped in PRs #2199 and #2265. This PR restores the subset that can pass without any infrastructure or Canvas API changes.

## Checklist

### DIFF
--- a/e2e/api/tests/plugin.api.spec.ts
+++ b/e2e/api/tests/plugin.api.spec.ts
@@ -48,38 +48,17 @@ test.describe("Plugin install/uninstall lifecycle via API", () => {
     sceneId = sc.createScene.scene.id;
   });
 
-  test.fixme(
-    "Uninstall and re-install the reearth plugin",
-    // Server returns "not found" when uninstalling the built-in reearth plugin.
-    // The built-in plugin cannot be managed via install/uninstall mutations.
-    async ({ gqlClient }) => {
-      const { status: uninstallStatus, data: uninstallData } =
-        await gqlClient.mutate<{
-          uninstallPlugin: {
-            pluginId: string;
-            scene: { id: string };
-          };
-        }>(UNINSTALL_PLUGIN, {
-          input: { sceneId, pluginId: "reearth" }
-        });
-
-      expect(uninstallStatus).toBe(200);
-      expect(uninstallData.uninstallPlugin.pluginId).toBe("reearth");
-
-      const { status: installStatus, data: installData } =
-        await gqlClient.mutate<{
-          installPlugin: {
-            scene: { id: string };
-            scenePlugin: { pluginId: string };
-          };
-        }>(INSTALL_PLUGIN, {
-          input: { sceneId, pluginId: "reearth" }
-        });
-
-      expect(installStatus).toBe(200);
-      expect(installData.installPlugin.scenePlugin.pluginId).toBe("reearth");
-    }
-  );
+  test("Built-in reearth plugin cannot be uninstalled via API", async ({
+    gqlClient
+  }) => {
+    // The server intentionally blocks managing built-in plugins.
+    // Verify this constraint is enforced.
+    await expect(
+      gqlClient.mutate(UNINSTALL_PLUGIN, {
+        input: { sceneId, pluginId: "reearth" }
+      })
+    ).rejects.toThrow();
+  });
 });
 
 // Negative scenarios
--- a/e2e/api/tests/plugin.api.spec.ts
+++ b/e2e/api/tests/plugin.api.spec.ts
@@ -52,12 +52,12 @@ test.describe("Plugin install/uninstall lifecycle via API", () => {
     gqlClient
   }) => {
     // The server intentionally blocks managing built-in plugins.
-    // Verify this constraint is enforced.
+    // Verify this constraint is enforced — the server returns "not found" for system plugins.
     await expect(
       gqlClient.mutate(UNINSTALL_PLUGIN, {
         input: { sceneId, pluginId: "reearth" }
       })
-    ).rejects.toThrow();
+    ).rejects.toThrow("not found");
   });
 });
 
--- a/e2e/pages/projectScreenPage.ts
+++ b/e2e/pages/projectScreenPage.ts
@@ -54,7 +54,7 @@ export class ProjectScreenPage {
   cancelLayerStyleButton: Locator;
 
   constructor(private page: Page) {
-    this.scenePanel = this.page.getByText("Scene");
+    this.scenePanel = this.page.locator('[aria-expanded]:has-text("Scene")').first();
     this.sceneItems = this.page.locator(
       '[data-testid="editor-map-scene-item"]'
     );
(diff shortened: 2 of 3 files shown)


e2e/tests/project/projects.spec.ts
```diff
@@ -314,13 +314,12 @@ test.describe("Project Management", () => {
...
+  test("Should navigate to Scene panel and verify scene items", async () => {
     test.info().annotations.push({ type: "story", description: "US-EDIT-003" });
     test.setTimeout(60000);
-    await projectScreen.scenePanel.click();
-    await page.waitForTimeout(1000);
-
-    await expect(projectScreen.getSceneItemByName("Main")).toBeVisible();
+    // The Scene panel is always open by default in the left column.
+    // Clicking its header collapses it, so we assert visibility directly.
+    await expect(projectScreen.getSceneItemByName("Main")).toBeVisible({ timeout: 15000 });
```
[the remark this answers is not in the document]
ZTongci: Fixed. The scenePanel locator is now changed to use the aria-expanded attribute on the Collapse header. The test checks if aria-expanded is false before clicking to expand, so it works whether localStorage has a collapsed state or not.

e2e/pages/projectScreenPage.ts
```diff
@@ -145,9 +145,7 @@ export class ProjectScreenPage {
   }
 
   getSceneItemByName = (name: string): Locator =>
-    this.page.locator(
-      `[data-testid="editor-map-scene-item"] div:has-text("${name}")`
-    );
+    this.page.locator(`[data-testid="editor-map-scene-item"]:has-text("${name}")`);
```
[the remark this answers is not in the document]
ZTongci: Fixed. Changed to use getByTestId('editor-map-scene-item').filter({ hasText: name }), which is the same pattern already used by getLayerByName. This avoids string interpolation into CSS selectors and uses Playwright's built-in text filtering.

e2e/api/tests/plugin.api.spec.ts
```diff
@@ -48,38 +48,17 @@ test.describe("Plugin install/uninstall lifecycle via API", () => {
...
+    gqlClient
+  }) => {
+    // The server intentionally blocks managing built-in plugins.
+    // Verify this constraint is enforced.
+    await expect(
+      gqlClient.mutate(UNINSTALL_PLUGIN, {
+        input: { sceneId, pluginId: "reearth" }
+      })
+    ).rejects.toThrow();
+  });
```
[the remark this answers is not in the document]
ZTongci: The gqlClient.mutate() function throws specifically when body.errors?.length is truthy (see api/graphql/client.ts). A 401 or network error produces a different response shape that does not have a GraphQL errors array, so it would not trigger the same throw path. The rejects.toThrow() assertion is sufficient here. Keeping it as is.

e2e/tests/project/projects.spec.ts
```diff
@@ -314,13 +314,16 @@ test.describe("Project Management", () => {
...
     test.setTimeout(60000);
-    await projectScreen.scenePanel.click();
-    await page.waitForTimeout(1000);
-
-    await expect(projectScreen.getSceneItemByName("Main")).toBeVisible();
+    // localStorage may persist a collapsed state from a previous run.
+    // Expand the panel only if it is currently collapsed.
+    if ((await projectScreen.scenePanel.getAttribute("aria-expanded")) === "false") {
+      await projectScreen.scenePanel.click();
+      await page.waitForTimeout(500);
```
airslice: Could we rely on the following toBeVisible() auto-wait instead of waitForTimeout(500)? That would make the test less timing-dependent.

e2e/api/tests/plugin.api.spec.ts
```diff
@@ -48,38 +48,17 @@ test.describe("Plugin install/uninstall lifecycle via API", () => {
...
+  test("Built-in reearth plugin cannot be uninstalled via API", async ({
+    gqlClient
+  }) => {
+    // The server intentionally blocks managing built-in plugins.
+    // Verify this constraint is enforced.
+    await expect(
+      gqlClient.mutate(UNINSTALL_PLUGIN, {
+        input: { sceneId, pluginId: "reearth" }
+      })
+    ).rejects.toThrow();
```
airslice: Could we assert the expected error message/code here rather than only rejects.toThrow()?

airslice (approved):

e2e/tests/project/projects.spec.ts
ZTongci: Fixed. Removed the waitForTimeout(500). The toBeVisible({ timeout: 15000 }) on the next line already waits for the item to appear, so the sleep was not needed.

e2e/api/tests/plugin.api.spec.ts
ZTongci: Fixed. Changed to rejects.toThrow("not found"). The server calls rerror.ErrNotFound when the plugin ID is a system plugin (pid.System() == true). This error message is distinct from auth errors, which throw with "GraphQL request failed (401): ...".

airslice (approved):

PR #2371, opened 2026-08-05
fix(server): stop scene lock release from getting stuck on a canceled context (REL-03)

wilfredmulenga:
REL-03 from the compliance scan flags that ReleaseSceneLock discards SaveLock's error and reuses the same request context the failed publish was using. If that context is already canceled (client disconnected mid upload), the SaveLock write can fail silently and the scene lock stays stuck at LockModePublishing, blocking every later mutation on that scene (layer/property/style edits, further publishes, even deletion) until the next process restart.

Detaches the context before saving the lock, bounded by a short timeout, and logs the error instead of swallowing it. Same pattern already used elsewhere in this codebase for a status write that needs to survive a canceled parent context.

Added a regression test that fails without the fix (SaveLock receiving an already-canceled context) and passes with it.

### DIFF
--- a/server/internal/usecase/interactor/common.go
+++ b/server/internal/usecase/interactor/common.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"net/url"
 	"strings"
+	"time"
 
 	accountsGateway "github.com/reearth/reearth-accounts/server/pkg/gateway"
 	accountsID "github.com/reearth/reearth-accounts/server/pkg/id"
@@ -18,6 +19,7 @@ import (
 	"github.com/reearth/reearth/server/pkg/project"
 	"github.com/reearth/reearth/server/pkg/scene"
 
+	"github.com/reearth/reearthx/log"
 	"github.com/reearth/reearthx/rerror"
 	"github.com/reearth/reearthx/usecasex"
 )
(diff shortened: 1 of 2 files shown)


airslice (approved):
LGTM!

server/internal/usecase/interactor/common_test.go
```diff
@@ -0,0 +1,50 @@
...
+	repo := &recordingSceneLockRepo{}
+	d := commonSceneLock{sceneLockRepo: repo}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	d.ReleaseSceneLock(ctx, id.NewSceneID())
+
+	assert.NoError(t, repo.saveLockCtxErr)
+}
```
ZTongci, line 54: saveLockCtxErr starts as nil (Go zero value). If a future change removes or skips the SaveLock call, this assert.NoError still passes because nil == no error. The test does not check that SaveLock was actually called. Adding a called bool field to the mock and asserting assert.True(t, repo.called) would catch that case.

ZTongci (approved):

mkumbobeaty (approved):

server/internal/usecase/interactor/common.go
```diff
@@ -176,6 +176,14 @@ func (i commonSceneLock) UpdateSceneLock(ctx context.Context, s id.SceneID, befo
 	return nil
 }
 
+// ReleaseSceneLock discards SaveLock's error and reuses the caller's context
+// as-is (compliance scan REL-03). If the caller's context is canceled (e.g. a
+// publish request whose client disconnected mid upload) the SaveLock write
```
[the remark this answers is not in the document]
wilfredmulenga: Fixed, hyphenated to "mid-upload".

server/internal/usecase/interactor/common_test.go
wilfredmulenga: Good catch. Added a `saveLockCalled` flag on the mock and asserted it. Verified it catches exactly this: temporarily stubbed `ReleaseSceneLock` to do nothing (skip the SaveLock call entirely), and the test now fails as expected -- previously it would have passed since `saveLockCtxErr` stays at its nil zero value.

airslice (approved):

Issue #164, opened 2021-11-05
Background color of markdown text block is not transparent


rot1024:
**Describe the bug**

Background color of markdown text block becomes white (light mdoe) and black (dark mode)

This unexpected behavior is caused by update of github-markdown-css

**To Reproduce**

Add a text block to the infobox and enable "Markdown"

**Expected behavior**

Background color should be transparent

**Screenshots**

![image](https://user-images.githubusercontent.com/3888696/140452918-f731d519-31cd-4ae6-a0b2-792c5822c233.png)

PR #2373, opened 2026-08-07
feat(web,server): allow users to toggle water mask rendering on terrain

ZTongci:
## Summary
- Adds a **Water Mask** on/off switch in the Editor terrain settings panel, directly below the terrain enable toggle
- Toggle is visible only when terrain is enabled (\`availableIf: terrain = true\`)
- Defaults to \`false\` — existing scenes without the field set will have water mask disabled; users can enable it from the UI

## Changes
- \`server/pkg/builtin/manifest.yml\` / \`manifest_ja.yml\`: new \`terrainWaterMask\` bool field in both terrain schema groups
- \`web/src/app/types/sceneProperty.ts\`: \`terrainWaterMask?: boolean\` added to \`TerrainProperty\`
- \`web/src/app/utils/convert-object.ts\`: maps \`terrain.terrainWaterMask\` → \`terrain.waterMask\` (passed to \`@reearth/core\`)
- \`web/package.json\`: bumps \`@reearth/core\` to \`0.0.7-alpha.79\`

## Core dependency
[reearth/core#164](https://github.com/reearth/core/pull/164) — \`feat(terrain): add configurable waterMask prop\`
Published as \`@reearth/core@0.0.7-alpha.79\`.

https://github.com/user-attachments/assets/4ac6aace-f64b-47f1-9d0d-9dc228289266


server/pkg/builtin/manifest.yml
```diff
@@ -397,6 +406,11 @@ extensions:
               type: bool
               title: Terrain
               description: Show elevation when close to the surface.
+            - id: terrainWaterMask
+              type: bool
+              title: Water Mask
+              description: Renders water bodies with a realistic water surface effect. Enabled by default.
+              defaultValue: true
```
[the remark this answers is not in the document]
ZTongci: Fixed. Added availableIf: terrain = true to terrainWaterMask in the legacy cesium schema group. It was already in the cesium-beta group but was missing in the legacy one.

web/src/app/utils/convert-object.ts
```diff
@@ -87,6 +87,7 @@ export const sceneProperty2ViewerPropertyMapping: Mapping = {
     }
   ],
   "terrain.terrain": "terrain.enabled",
+  "terrain.terrainWaterMask": "terrain.waterMask",
   "terrain.terrainType": "terrain.type",
```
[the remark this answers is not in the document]
ZTongci: The core version bump will be handled separately once PR #164 in the core repo is merged and released. The manifest changes and the convert-object mapping work on their own. In local dev we use link:../../core so the version does not matter there.

server/pkg/builtin/manifest.yml
```diff
@@ -397,6 +406,15 @@ extensions:
...
               description: Show elevation when close to the surface.
+            - id: terrainWaterMask
+              type: bool
+              title: Water Mask
+              description: Renders water bodies with a realistic water surface effect.
+              defaultValue: false
+              availableIf:
+                field: terrain
+                type: bool
+                value: true
```
airslice: We don't need to update this, this is under `cesium`, which is not used by Visualizer.

server/pkg/builtin/manifest_ja.yml
```diff
@@ -204,6 +207,9 @@ extensions:
           terrain:
             title: 地形
             description: 有効にすると、標高データが読み込みこまれ、立体的な地形を表現することができます。
+          terrainWaterMask:
+            title: ウォーターマスク
+            description: 水域をリアルな水面エフェクトで表示します。
```
airslice: We don't need this for the same reason.

server/pkg/builtin/manifest.yml
ZTongci: Removed. The terrainWaterMask field and its availableIf condition are now deleted from the cesium group.

server/pkg/builtin/manifest_ja.yml
ZTongci: Removed.

airslice (approved):

PR #2385, opened 2026-08-19
refactor(web): improve Recycle Bin and workspace member workflows [VIZ-DEV-179]

mkumbobeaty:
# Overview
This PR improves the Recycle Bin and Members flows by fixing several edge cases around mutations and UI state.
## What I've done
- Improve error handling for project and member mutations.
- Prevent incorrect cache updates when mutations fail.
- Prevent duplicate submissions while requests are in progress.
- Fix last workspace owner validation.
- Preserve member search state after updates.
- Add tests for the updated behaviors.
## Checklist


wilfredmulenga (commented):
Thanks for the PR. I found one small issue. There is a small bug which is hard to picture in real usage but might exist.

Its possible for the recycle bin to list everything twice with this change. To see it: open the bin, archive another project from a duplicate tab, then move back to the first tab and use the sidebar to go to Projects and back (reloading the page hides it). On main branch, it shows 16 items, but after the change to add the `cache-and-network fetch policy` it will show 32. Looks like the cache-and-network fetch policy is the cause. The deleted projects query uses a merge function that appends results so that paging works, and the refetch() it replaces overwrote them instead. 

Reverting that one line should sort it. But please let me know if you would prefer to keep this change.

airslice:
@mkumbobeaty Please check @wilfredmulenga 's comment, other parts LGTM! Thank you!

airslice (approved):

PR #2386, opened 2026-08-20
fix(server): report not found instead of crashing or returning null entries

wilfredmulenga:
## Problem

Two defects found while auditing the GraphQL error logs. Both are reached by asking for something that does not exist, which the API test suite does on purpose, so both fired on every test run and accounted for 64 ERROR lines in the dev project over the week to 2026-08-19.

**`uploadFileToProperty` crashed the request.** It dereferenced a nil entry returned by `Property.Fetch`. `Fetch` returns a slice with a nil entry for an ID that does not exist, so checking the length is not enough to know the item was found, and `pr[0].Scene()` panicked. gqlgen recovered it and reported `internal system error`. The scene and project lookups that follow had the same hole.

**`propertySchemas` and `plugins` returned a null element.** Both fields are non-null in the schema, `propertySchemas(id: [ID!]!): [PropertySchema!]!`, so a null entry for an unknown ID made gqlgen reject our own response with "the requested element is null which the schema does not allow".

## What this changes

Both now report not found, which is what the single-item equivalents already do.

## Verification

The two API tests that cover these paths, "Cannot upload file to a non-existent property" and "propertySchemas: non-existent schemas throws error", still pass, and the panic and the schema violation no longer appear in the server log. `go test ./...` passes from `server/`.

## Relationship to the logging PR

Split out of the error-severity PR on review feedback, since these are behaviour fixes rather than logging changes. They are independent: this one can merge in either order. Once both are in, these paths answer not found and are logged at WARN rather than ERROR.

Task: VIZ-DEV-81


airslice (approved):
LGTM! Thank you!

PR #2387, opened 2026-08-20
fix(web): reduce redundant GetScene refetches when adding system tiles

mkumbobeaty:
# Overview
This PR optimizes system-tile creation for Street View and Google Map Search by reducing redundant GetScene refetches.

Previously, addSystemTile performed an explicit refetch() followed by several sequential property mutations. Since addWidget, addPropertyItem, and updatePropertyValue already trigger GetScene refetches, installing a single system tile could result in roughly five full-scene fetches.

Because GetScene payload and server-side assembly costs grow with project size, these repeated refetches can significantly increase network usage and user-visible latency on larger projects.

This change avoids unnecessary intermediate refetches and ensures the scene is refreshed only when needed after the system-tile setup is complete.
## Checklist


airslice (approved):
LGTM! Thank you!

PR #2388, opened 2026-08-21
fix(server): do not panic importing an export that has no story

wilfredmulenga:
## Problem

`story` is `omitempty` in the export format, so a project exported without one has no `story` key and `sceneJSON.Story` is nil. `ImportStory` ranged over it and panicked. The worker recovers it as `panic during import: ... nil pointer dereference` and sets `importStatus: FAILED`, but the project, assets, scene and layers are already written, leaving a half-imported project.

Unnoticed because the web app and the internal API both create a default story on project creation. A project created via the GraphQL API without `createStory` exports fine and then cannot be imported.

## Fix

Return an empty result when there is no story to import.

## Verification

New unit test panics without the guard, passes with it. Export and re-import of a story-less project through `/api/split-import` now reaches `importStatus: SUCCESS`. `go test ./...` and `golangci-lint run --new-from-rev=origin/main` clean.

The e2e spec covering that round trip is a separate PR and should merge after this one.

Task: VIZ-DEV-81


airslice (approved):
LGTM! Thank you!

PR #2390, opened 2026-08-21
fix(server,web): stop a member search that matches nobody from raising an error

wilfredmulenga:
# Overview

Searching the invite-member box for someone who is not registered shows a red error notification containing a raw internal message, rather than the modal's own "user not found" state.

`UserLoader.SearchUser` calls the accounts client's `FindByAlias` and returns its error unchanged. A missing record comes back as an error rather than an empty result, so a normal "nobody matched" reaches the web client as a GraphQL error, and the global error link turns it into an error notification.

The box also told people to search by email address or user name, but the lookup matches on alias only, so both of the fields it named were the wrong thing to type.

## What I've done

- Treat a missing record from `FindByAlias` as an empty result instead of an error, and let every other error propagate as before
- Change the modal copy and placeholder to say alias, which is what the lookup actually matches
- Disable Add until the search has found somebody, so there is nothing to click while the list is empty
- Add tests for all of the above

The accounts client only gives a typed error for 401, so the missing-record case is matched on its message. That check is applied to a single lookup by alias, where "not found" can only mean the alias matched nobody.

## What I haven't done

The server still logs one ERROR line per miss. That log is written inside the accounts client before it returns the error, so it cannot be suppressed from here. Removing it needs a change in reearth-accounts.

Left the lookup itself alone. The accounts client also exposes `FindByNameOrEmail`, so searching by email could be made to work, but that is a behaviour change rather than a fix.

## How I tested

Server: `go test ./internal/adapter/gql/` passes, including the pre-existing loader test. Web: three new tests pass, and `tsc` and `eslint` are clean.

Each fix was checked by reverting it and confirming the matching test failed. The rest of `go test ./...` is unchanged from main, where the `e2e` package already fails for unrelated reasons.

## Memo

Touches `AddMemberModal.tsx`, which PR 2385 also edits, so one of the two will need a manual merge.

## Checklist

---

A member search that matched nobody surfaced the accounts client's raw "not found" as an error notification; now returns an empty result so the invite modal shows its own "user not found" state, with the search box corrected to say it matches an alias.

VIZ-DEV-81


server/internal/adapter/gql/loader_user.go
```diff
@@ -84,6 +87,22 @@ func (c *UserLoader) SearchUser(ctx context.Context, nameOrEmail string) (*gqlmo
 	return nil, nil
 }
 
+// isAccountsNotFound reports whether err is the accounts API saying the record
+// does not exist, as opposed to a real failure.
+//
+// Searching for someone who has not signed up is a normal outcome, but the
+// accounts client reports it as an error rather than an empty result, and the
+// caller has no way to tell the two apart: only 401 gets a typed error
+// (gqlerror.ErrUnauthorized), everything else is returned verbatim. Left as an
```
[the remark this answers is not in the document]
wilfredmulenga: Reworded in cc848a2ad, though for a slightly different reason than the one given.

`ErrUnauthorized` does exist, in the accounts dependency at `pkg/gqlclient/gqlerror/error.go:14`. The real problem is that `gqlerror` is ambiguous in this repo: `internal/app/graphql.go` uses that name for `vektah/gqlparser/v2/gqlerror`, and `auth_client.go` imports the accounts one under the alias `accountsGqlError`. So a bare `gqlerror.ErrUnauthorized` reads as the vektah package, where it genuinely is not defined.

The comment now names the package path instead of a bare qualifier.

server/internal/adapter/gql/loader_user.go
```diff
@@ -84,6 +87,22 @@ func (c *UserLoader) SearchUser(ctx context.Context, nameOrEmail string) (*gqlmo
...
+// gives no way to tell the two apart: it returns a typed error only for 401
+// (ErrUnauthorized, in the reearth-accounts package pkg/gqlclient/gqlerror),
+// and everything else verbatim. Left as an error it reaches the web client as
+// a GraphQL error, which the global error link turns into an error
+// notification carrying the raw message. Matching on the message is the only
+// handle available; it is narrow because it is applied to a single lookup by
+// alias, where "not found" can only mean the alias matched nobody.
+func isAccountsNotFound(err error) bool {
+	return err != nil && strings.Contains(strings.ToLower(err.Error()), "not found")
+}
```
airslice, line 104: Will this accidiently treat any error who contains the word `not found` as account not found?

airslice (approved):
Basically LGTM! I left one comment, please help check it, thank you!

server/internal/adapter/gql/loader_user.go
wilfredmulenga: Good catch — it could, as written. Fixed in cea243d6e: instead of my own whole-message substring match, it now reuses the accounts client's own ErrUserNotFound predicate, which type-asserts to graphql.Errors first and only then checks the message. So a non-GraphQL error that merely contains "not found" (e.g. a transport 404) is no longer swallowed — only a genuine accounts not-found is. Added a regression test (TestUserLoader_SearchUser_PlainNotFoundStringPropagates) for exactly that case, and made the not-found test return a realistic graphql.Errors like production.

airslice (approved):

PR #2400, opened 2026-08-31
docs(server): note accepted tradeoff for in-memory export zip assembly (SCA-07)

wilfredmulenga:
## Overview

The project export zip is assembled in memory, so peak memory scales with the export size.

This is a docs-only change: an accepted-tradeoff note at `maxExportZipBytes` so the reasoning lives in the code (and is visible to the compliance scan), not only in the tracker. No behaviour change.

## Rationale

At current load this is a dormant, low-risk concern, and the change it would take to address it is not justified today. It is documented in code with a revisit trigger; the analysis and the options are tracked internally.

## Checklist


airslice (approved):
LGTM!

PR #2401, opened 2026-08-31
fix(server): record terminal status and distinguish import-claim states (REL-08, REL-09)

wilfredmulenga:
## Overview

Two import-claim reliability fixes from the compliance scan, both in the claim path added by the REL-04 import cutover.

**REL-09** -- the split-upload worker returned on a `ClaimImport` error without writing any terminal status, while cleanup deleted the assembled upload, leaving the project stuck at `UPLOADING`. It now writes `FAILED` before returning.

**REL-08** -- `ClaimImport` returned a bare bool that conflated "already succeeded" with "still processing", so a retryable (Pub/Sub) caller acked and dropped a redelivery whenever a claim was in progress -- an import whose worker died mid-run could be lost. `ClaimImport` now reports granted / already-succeeded / in-progress; handlers ack only on success and return a retryable error for in-progress.

## Tests

- REL-09: `ClaimImport` error path writes `FAILED` (verified it fails without the fix).
- REL-08: interactor reports in-progress vs already-succeeded correctly.
- Full `app`, `interactor`, and `mongo` suites pass.

## Checklist


airslice (approved):
One theoretical edge case: if writing the `FAILED` status fails after `ClaimImport` has already returned an error, the project could remain in `UPLOADING` after the assembled upload is cleaned up. This is unlikely, but logging the secondary update failure would make the condition easier to diagnose if it ever occurs.

wilfredmulenga:
Thanks — good edge case. The secondary write is already logged inside UpdateImportStatus, but only via a default-severity Printf, which is easy to miss when diagnosing a stuck import. Raised it to log.Errorfc with the project id and target status in b040de94c, so if the FAILED-status write ever fails and leaves the project in UPLOADING, it surfaces at ERROR with enough context to trace. Kept it to logging rather than returning the error up, since every UpdateImportStatus call site ignores the return and this path is rare.

airslice (approved):

PR #2402, opened 2026-08-31
fix(server): guard findOne against an empty filtered result (REL-10)

wilfredmulenga:
## Overview

`findOne` in the style, scene, property, and project mongo repos returned `c.Result[0]` directly. `FindOne` can return no error while the readable/scene/workspace filter drops the matched document, leaving `Result` empty, so the index panicked (a 500) instead of reporting not-found. This applies the same length guard already merged for `storytelling.findOne`.

## Tests

- A representative regression test on the style repo: a scene-scoped read of an out-of-scene document returns not-found instead of panicking (verified it panics without the guard). The other three repos take the identical guard.
- Full `mongo` suite passes.

## Checklist


airslice (approved):
LGTM, thank you!

PR #2403, opened 2026-08-31
fix(server): report a missing scene as not found, not operation denied

wilfredmulenga:
Six mutations (`AddLayerSimple`, `AddStyle`, `CreateStory`, `CreatePage`, `UpdatePage`, `RemovePage`) called `CanWriteScene` on a client-supplied scene id before confirming the scene exists, so a missing scene came back as `operation denied` instead of `not found`. Each now looks the scene up first, matching `AddWidget`.

Scenes that exist are unaffected, so the SEC-01 / VIZ-DEV-98/99 permission tests still hold. Two fall-outs: `CreateStory` no longer writes a property before the scene is confirmed, and nlslayer drops a copy-pasted `validateGeoJSONFeatureCollection` log message.

Six tests cover the missing-scene case. `go test ./...` and `golangci-lint --new-from-rev=origin/main` clean.

Task: VIZ-DEV-81

### DIFF
--- a/server/internal/usecase/interactor/nlslayer.go
+++ b/server/internal/usecase/interactor/nlslayer.go
@@ -118,8 +118,15 @@ func (i *NLSLayer) AddLayerSimple(ctx context.Context, inp interfaces.AddNLSLaye
 		}
 	}()
 
+	// Look up the scene first so a scene that does not exist is reported as not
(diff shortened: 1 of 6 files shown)


airslice (approved):
LGTM! Thank you!

PR #2423, opened 2026-09-17
fix(server): batch buffer rewrites during asset import instead of one pass per asset

wilfredmulenga:
## Problem
Importing a project rewrites the whole project.json buffer once per imported asset, allocating and copying the entire buffer on every iteration even when there are no matches.

## Fix
Accumulate all ID replacements during the loop and apply them in a single batched pass afterward, using the same helper already used for layers, styles, and scenes.

## Verification
- Added a regression test importing multiple assets and confirming every asset's URL is rewritten, not just the last one
- go test ./... passes


wilfredmulenga:
Closing — keeping this locally for now to avoid extra CI/alert noise while the other rollout PRs are still in flight.

PR #2425, opened 2026-09-17
fix(server): close each zip entry reader at the end of its own iteration

wilfredmulenga:
## Problem
Asset import opens a reader for each asset's zip entry and defers closing it, but the defer lives inside the per-asset loop body. Every deferred close accumulates and only runs once the whole import finishes, so every entry's reader (and its retained decompression buffer) stays open for the entire import instead of being released after its own asset.

## Fix
Wrap each iteration's work in its own closure so the reader for that entry opens and closes within that same iteration, on every exit path.

## Verification
- Added a regression test confirming a failure partway through import still stops immediately with that error and doesn't fabricate a result for the failed asset, guarding against the kind of error-propagation mistake this restructuring can introduce
- go test ./... passes


wilfredmulenga:
Closing — keeping this locally for now to avoid extra CI/alert noise while the other rollout PRs are still in flight.

